PR #7, opened 2024-12-03
DEV: Prepare main for 25.01 developement

seberg:
This updates our and legate/cupynumeric versions to 25.01 (development).

The biggest follow-up tasks to do soon(ish) is that we should update to RAPID 24.10 including rapids-cmake 24.10.
IIUC, this means that we have to switch to `scikit-build-core`, which may be some work.


VERSION
```diff
@@ -1 +1 @@
-24.11.00
+25.01.00.dev
```
jameslamb, line 1: Including the `.dev` isn't strictly necessary, but I guess it doesn't hurt. The content of this file should only affect the versions for locally-built things, since all the CI scripts overwrite this file's contents with the output of `rapids-generate-version` (which reads from tags.

jameslamb (approved):
Nice! Remember to add `[skip ci]` to the title of the PR (or immediately, manually cancel the automatically-triggered build) if you don't want this to become part of one last `24.11.*` nightly.

PR #8, opened 2024-12-03
Update to use Rapids 24.10 at runtime and build setup

seberg:
The main change here is the build setup @jameslamb.

The `pylibcudf` changes were started by Matt before, the C change, is just a very simple change in public API (there is the new stream parameter).

*~Draft, since the update to 25.01.00.dev should happen first~ Ready for review, but I suspect tests will fail due to a small (fixed!) legate issue for another day*.


conda/recipes/legate-dataframe/meta.yaml
```diff
@@ -89,7 +89,7 @@ test:
   # import tests at build time do not work for the CUDA packages,
   # because builds happen on machines without a GPU.  So no tests.
   commands:
-    - pip show legate-dataframe
+    - pip show legate-dataframe-cu12
```
jameslamb, line 92: This `-cu{11,12}` convention shouldn't be used for conda packages, since conda has a better solution for differentiating between CUDA versions (the `__cuda` virtual package).

To prevent these suffixes from making it into conda packages, across RAPIDS we pass this to `pip install` in conda builds:

```text
--config-settings rapidsai.disable-cuda=true
```

Example from `cudf`: https://github.com/rapidsai/cudf/blob/1b82963df736f3ad71b003443a4de1414f3ce2e5/build.sh#L78

I recommend doing the same here.

dependencies.yaml
```diff
@@ -92,6 +99,18 @@ dependencies:
             packages:
               - cuda-nvcc
 
+  rapids_build_skbuild:
+    common:
+      - output_types: [conda, requirements, pyproject]
+        packages:
+          - rapids-build-backend>=0.3.0,<0.4.0.dev0
```
jameslamb, line 106:
```suggestion
          - rapids-build-backend>=0.3.2,<0.4.0.dev0
```

Since this support is being added brand-new here, and since this project's build tool dependencies don't affect any other RAPIDS projects, I recommend going all the way to the latest version ([`0.3.2`](https://pypi.org/project/rapids-build-backend/#history)))

conda/recipes/legate-dataframe/meta.yaml
```diff
@@ -61,14 +61,14 @@ requirements:
     # is compatible with an existing cupynumeric version.
     # (may also stabilize not using debug/sanitizer builds)
     - cupynumeric
-    - cython>=3.0.0
     - python
-    - pip
```
jameslamb, line 66: I think you still need `pip` as a host dependency, because the build ends up invoking it:

https://github.com/rapidsai/legate-dataframe/blob/dd9bc3ccd4a6e70d5ef8ac78ca525534a592f3d9/build.sh#L158

From the logs ([link](https://github.com/rapidsai/legate-dataframe/actions/runs/12178643194/job/33969113140?pr=8#step:5:266)) it looks like you're still getting it, but maybe only transitively as a dependency of something else. It'd be better to specify it explicitly, to prevent surprises from that changing in the future (as recently happened to some projects with `setuptools`, e.g. https://github.com/rapidsai/dependency-file-generator/pull/113#issuecomment-2402855030).

jameslamb (approved):
Left some small suggestions, but overall this is great!!!

I'm really happy to see the switch to `scikit-build-core` and `rapids-build-backend`, and glad that seems like it went relatively smoothly.

seberg:
OK, compatible legate packages are available now, so things are passing.  The last commit just fixes to use `void legate::start()` since the other is `[[deprecated]]` (it actually failed the build.
That seems simple enough, so going to put this in.

PR #10, opened 2024-12-23
remove versioneer references, update to latest rapids-dependency-file-generator

jameslamb:
## Description

Just some minor changes:

* removes configuration related to `versioneer` (not used in this project)
* removes an unused YAML anchor in `dependencies.yaml`
* updates to latest `rapids-dependency-file-generator` (v1.17.0) in pre-commit hooks 

## Checklist
- [ ] Run `./build.sh test` for local testing, see [CONTRIBUTING.md](https://github.com/rapidsai/legate-dataframe/blob/main/CONTRIBUTING.md#testing).


dependencies.yaml
```diff
@@ -188,7 +188,7 @@ dependencies:
       - output_types: conda
         packages:
           - &cudf_unsuffixed cudf==24.10.*,>=0.0.0a0
-          - &pylibcudf_unsuffixed pylibcudf==24.10.*,>=0.0.0a0
+          - pylibcudf==24.10.*,>=0.0.0a0
```
seberg, line 191: Hmmm, I don't quite understand what the:
```
 - {matrix: null, packages: [*cudf_unsuffixed]}
```
does below right now 🤔?

seberg (approved):
Looks good, thanks.  I don't quite understand the `cudf_unsuffixed` use right away, but it doesn't matter for putting this in either way.

dependencies.yaml
jameslamb: That "fallback matrix" (the one `rapids-dependency-file-generator` hits when none of the others matches) is what populates `pyproject.toml` in source control when the `rapids-dependency-file-generator` pre-commit hook runs.

That list in `pyproject.toml` isn't every *used* for anything, because `rapids-build-backend` will get a CUDA-major-version-specific list at build time. But it's helpful documentation.

In RAPIDS, by convention we have this be a list of dependencies without CUDA suffixes. But there are some reasons for that convention in RAPIDS that don't apply here:

* RAPIDS supports two major versions of CUDA
* RAPIDS libraries are sometimes built in contexts where all the packages have CUDA suffixes removed (like https://github.com/rapidsai/devcontainers)

If you'd prefer, I can remove all this `cuda_suffixed` stuff here for simplicity, and we could consider re-introducing it in the future if this project starts supporting multiple major versions of CUDA.

jameslamb: By the way, this conversation made me realize that removing the anchor here wasn't the right choice... it's actually better to *add* `pylibcudf` to the fallback matrix, so you can tell by looking at `pyproject.toml` that `legate-dataframe` depends on it. Did that here: https://github.com/rapidsai/devcontainers

seberg: Thanks, I guess I was more confused by the what than the why about the difference :).  Let's put it in!

PR #11, opened 2025-01-09
MAINT: Quick fix for the one pool world

seberg:
This is just Wonchan's old PR with small comments added.  I plan to open an issue about actually adding at least some estimates about memory usage. This may not matter for many simple pipelines, but also could matter a lot for some.

(Memory use is not easy to bound, e.g. for joins or csv reading, although even a very high overestimate may be useful if we read a small csv for example.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,9 +27,9 @@ RUN mkdir -p /opt/legate-dataframe/conda-env-file
 COPY ./conda/environments/*.yaml /opt/legate-dataframe/conda-env-file/
 
 # To ensure we find the GPU version of legate in the docker build.
-ARG CONDA_OVERRIDE_CUDA=$CUDA_VERSION
+ARG CONDA_OVERRIDE_CUDA=12.4
 RUN /bin/bash -c '/opt/conda/bin/mamba env create --name legate-dev --file \
-  /opt/legate-dataframe/conda-env-file/all_cuda-$(cut --output-delimiter="" -d "." -f 1,2 <<< ${CUDA_VERSION})_arch-x86_64.yaml'
+  /opt/legate-dataframe/conda-env-file/all_cuda-$(cut --output-delimiter="" -d "." -f 1,2 <<< ${CONDA_OVERRIDE_CUDA})_arch-x86_64.yaml'
 
 # Build and install legate-dataframe
 WORKDIR /opt/legate-dataframe
--- a/ci/run_pytests.sh
+++ b/ci/run_pytests.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2024, NVIDIA CORPORATION.
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
 
 # [description]
 #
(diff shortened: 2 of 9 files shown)


cpp/src/core/library.cpp
```diff
@@ -69,6 +69,47 @@ class Mapper : public legate::mapping::Mapper {
...
+          auto nrank    = task.get_launch_domain().get_volume();
+
+          // Space for the exchange buffers containing size chunks:
+          size_t size_exchange_nbytes = nrank * nrank * 2 * sizeof(std::size_t);
+          // Space for column packing metadata exchange
+          size_t sizeof_serialized_column = 6 * 8;  // not public, rough upper bound
+          // num_cols * 2 for the string column child and + 1 for the number of cols
+          size_t metadata_nbytes = nrank * (num_cols * 2 + 1) * sizeof_serialized_column;
+
+          return size_exchange_nbytes + metadata_nbytes;
```
seberg, line 102: Not sure this is really worthwhile, since the initial PR had the `return 0` here, I thought I would get the correct value.

For aggregations that is nice, for `join` would have to fetch both tables, although that shouldn't be too bad (we just don't have a convenience function to do that via the task).

Locally, I see cuda allocation failures in some join tasks, it isn't clear to me yet why those happen, though.

Dockerfile
```diff
@@ -27,9 +27,9 @@ RUN mkdir -p /opt/legate-dataframe/conda-env-file
 COPY ./conda/environments/*.yaml /opt/legate-dataframe/conda-env-file/
 
 # To ensure we find the GPU version of legate in the docker build.
-ARG CONDA_OVERRIDE_CUDA=$CUDA_VERSION
+ARG CONDA_OVERRIDE_CUDA=12.4
 RUN /bin/bash -c '/opt/conda/bin/mamba env create --name legate-dev --file \
-  /opt/legate-dataframe/conda-env-file/all_cuda-$(cut --output-delimiter="" -d "." -f 1,2 <<< ${CUDA_VERSION})_arch-x86_64.yaml'
+  /opt/legate-dataframe/conda-env-file/all_cuda-$(cut --output-delimiter="" -d "." -f 1,2 <<< ${CONDA_OVERRIDE_CUDA})_arch-x86_64.yaml'
```
seberg, line 32: Just some smaller changes so that `./ci/manual_test.sh` at least runs again (hand't really used it in a while).

ci/run_pytests.sh
```diff
@@ -26,6 +26,7 @@ cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../python/tests/
 LEGATE_TEST=${LEGATE_TEST:-1} \
 legate \
     --gpus "$(nvidia-smi -L | wc -l)"\
+    --fbmem=4000 \
```
seberg, line 29: Somehow, it seems legate is reserving almost all available GPU memory now, and that just doesn't work for us...

danielfrg: @magnatelee does that sound right to you?

magnatelee: yeah like @manopapad said on slack, the default behavior now tries to reserve the entire available GPU memory. we'll be transitioning to a mode where things are allocated incrementally.

RAMitchell (approved):

jameslamb (approved):

PR #15, opened 2025-01-22
API,MAINT: RAPIDS 24.12 and align DatetimeComponent with pylibcudf

seberg:
Legate bumped to using rapids-cmake 24.12, so let's use RAPIDS 24.12 also.

Most changes are just search/replace, but:
* This enables CI for Python 3.12, I forgot about that when we went to 24.10 (I think that was compatible).
* `libcudf` doesn't ship `libcudftestutil` anymore, instead we need to add it to our link targets (and the compile flags) to compile it here.
* **API change:** `(py)libcudf` now has `etract_component` and `DatetimeComponent`.  So align with that.  This means:
  * Python code needs to use the capitalizem, i.e. `DatetimeComponent.DAY` now.
  * `day_of_year` is a dedicated function in `(py)libcudf`, so it is gone for now.
  * (On the C-side, the same change would be needed.)

---

*~Unfortunately, the packages will probably be bad due to Legion linking issues.~*


ci/run_pytests.sh
```diff
@@ -18,15 +18,11 @@ set -e -E -u -o pipefail
 # Support invoking run_cudf_pytests.sh outside the script directory
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../python/tests/
 
-# Unless otherwise specified, use all available GPUs and set
+# Unless `LEGATE_CONFIG` is set, default to all available GPUs and set fbmem/sysmem.
 # LEGATE_TEST=1 to test broadcasting code paths (locally).
-# TODO: Set LEGATE_CONFIG instead (if undefined).  However,
-#       as of 2024-10-11 LEGATE_CONFIG seems broken:
-#       https://github.com/nv-legate/legate.core.internal/issues/1304
+LEGATE_CONFIG=${LEGATE_CONFIG:- --gpus="$(nvidia-smi -L | wc -l) --fbmem=4000 --sysmem=4000"} \
```
seberg, line 23: Admittedly, just part of the PR because the comment contains `24-10`.  But a simple improvement that works now: `LEGATE_CONFIG=` can be used to override settings.

madsbk (approved):
Looks good

jameslamb (approved):
Looks great, glad that we could add Python 3.12!

PR #17, opened 2025-01-29
ENH: Support more return columns

seberg:
Each column takes up some space in the return and this space is currently limited but settable.

This bumps the limit for all tasks so that we should be able to use well above 100 columns everywhere.

(Clearly, it is very plausible that we might bump this further if we run into a use-case that needs that.)

---

Tested just with csv, since that is where it came up (but if used, it tends to be the first call).  Either way, it should affect everything, so set it globally.

### DIFF
--- a/cpp/src/core/library.cpp
+++ b/cpp/src/core/library.cpp
@@ -122,7 +122,8 @@ legate::Library create_and_registrate_library()
     GlobalMemoryResource::set_as_default_mmr_resource();
   }
   // Set with_has_allocations globally since currently all tasks allocate (and libcudf may also)
-  auto options = legate::VariantOptions{}.with_has_allocations(true);
+  // Also ensure we can generally work with 100+ return columns.
+  auto options = legate::VariantOptions{}.with_has_allocations(true).with_return_size(8192);
   auto context =
     legate::Runtime::get_runtime()->find_or_create_library(library_name,
                                                            legate::ResourceConfig{},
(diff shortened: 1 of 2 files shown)


madsbk (approved):

RAMitchell (commented):
100 dataframe columns is not a lot for a ML workflow. Would it hurt to go higher e.g. up to 1000 or even 10000?

seberg:
I'll assume it doesn't really matter, so let me go to ~10x for no?  (if you don't have string columns, I think you actually get double, so it'll probably be ~5000).

seberg:
Hmmm, actually, it seems like just bumping is a problem:
> LEGION ERROR: Failed to reserve a dynamic memory pool of 4096 bytes for leaf task core::extract_scalar (UID 95574) in Z_COPY_MEM memory. You are actually out of memory here so you'll need to either allocate more memory for this kind of memory when you configure Realm which may necessitate finding a bigger machine. If you want to avoid finding out that you're out of memory this way you should instead use the 'MapperRuntime::acquire_pool' call to make sure that memory can be reserved for all pools in advance.

So, I am tempted to just put it in for now to unblock Artem, but we should see what is a good value (or if that z_copy_mem issue can be trivially avoided).

seberg:
Sorry, this is not a real problem, I just have to additionally increase the `--zcmem` when running the test.  And it doesn't look like this affects runs with few columns.

seberg:
Going to put this in.  The test covers only 1250, it works with 2500, but requires passing `--zcmem`.
But, I don't want to crash if that isn't set and I couldn't find an easy way to query what it is set to.

PR #18, opened 2025-01-29
WIP: Implement an experimental ibis backend for `legate-dataframe`

seberg:
This is to explorer `ibis` as a nicer Python API.  It is experimental for the time being (I hope naming is enough to clarify that), in that we might drop it and go another route.

This is largely working, and if it looks good to Gil then it might just need a bit of cleanup and maybe moving things around as well brief documentation.

It probably needs at least one pass by me, but I think it is ready for a first review.  Some notes:
* Supports most operations we have currently, except:
  * Time related operations
  * I skipped `replace_nulls`, because it seemed that `fill_nulls` also fills NaNs, so wanted to double check.
* Maybe `experimental_ibis` should be it's own package/module?  Does it matter for now?
* Currently it only executes the ibis IR directly.  E.g. the polars backend uses small re-writes for the `JoinChain` and `Aggregation`.
  Using small re-writes may be a bit nicer, although this doesn't seem too bad (column name handling is annoying, though).  We'll definitely want some re-writes in the future.

There are of course many things that won't work, and until a few more things are probably needed to be able to integrate the `ibis` tests.

**Breaking change:**
* I made `column.dtype()` return the `pylibcudf` dtype and kept `column.cudf_dtype()` for the old meaning, although I am not sure one should need it...

---

@gforsyth when you have time, maybe you can review?


python/legate_dataframe/experimental_ibis/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+            table = self._convert_object(obj)
+            if schema is None:
+                schema = infer_schema_from_logical_table(table)
+            else:
+                # Check if schema matches actual table (is it right to not
+                # enforce the schema by dropping cols or casting?)
+                actual_schema = infer_schema_from_logical_table(table)
+                if schema != actual_schema:
+                    # TODO: General Q: when to use Ibis errors and when not?!
+                    # (I find the broad use of IbisError a bit surprising.)
```
gforsyth, line 160: This (on the Ibis side) is ongoing work.

In the (very soon to be released) 10.0, we unified all of the various ways that different backends report failures under a common `TableNotFound` exception.  

But we do tend towards `IbisError` for user interactions, especially where passing through the backend error would lead to different exception types depending on the backend being used.

python/legate_dataframe/experimental_ibis/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+        if schema is not None:
+            schema = ibis.schema(schema)
+
+        if obj is not None:
+            table = self._convert_object(obj)
+            if schema is None:
+                schema = infer_schema_from_logical_table(table)
+            else:
+                # Check if schema matches actual table (is it right to not
+                # enforce the schema by dropping cols or casting?)
```
gforsyth, line 156: We try not to infer intent because it's a tricky problem.

1. w.r.t. column names, we definitely don't, because it's easy for the "wrong" schema to match a subset of columns and cause problems.  If users want to drop columns, this is what `mutate` and `select` are for.
2. w.r.t dtypes, this if more doable and we have a `castable` utility for determining lowest-common-denominator dtypes, but these decisions aren't always perfectly deterministic.

python/legate_dataframe/experimental_ibis/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+    def _to_ldf_table(
+        self,
+        expr: ir.Expr,
+        params: Mapping[ir.Expr, object] | None = None,
+        limit: int | Literal["default"] | None = None,
+        **kwargs: Any,
+    ) -> LogicalTable:
+        """Execute after ensuring that the result is a logical table."""
+        if limit is not None and limit != "default":
+            raise NotImplementedError("limit parameter not supported.")
```
gforsyth, line 212: When Ibis is used interactively, we run all commands eagerly by executing them with an implicit `LIMIT 11` at the end, so we can return the first 10 rows of everything (and an indicator ellipsis if there are >10 rows).

From my understanding of legate execution, this model doesn't really work, so the missing `limit` won't really be an issue.

python/legate_dataframe/experimental_ibis/schema.py
```diff
@@ -0,0 +1,64 @@
...
+}
+
+
+_to_plc_type = {i_t: DataType(plc_tn) for i_t, plc_tn in _to_plc_type_id.items()}
+
+_from_legate_df_type = {v: k for k, v in _to_plc_type.items()}
+
+
+def to_plc_type(ibis_type):
+    # TODO: Should we take nullability into account? (and if, how?)
```
gforsyth, line 38: If `legate` supports non-nullable columns, then maybe, because there are performance benefits, but if not, I would either ignore or error if someone asks for a non-nullable column / dtype

python/tests/test_ibis.py
```diff
@@ -0,0 +1,136 @@
...
+    # NOTE: As of writing, require schema.  In principle should infer schema
+    # and not actually read the full file immediately (i.e. make lazy).
+    table = b.read_csv(file, schema=schema)
+
+    sum_ab = (table.a + table.b).name("sum_ab")
+
+    res = sum_ab.execute()  # A columnar (pandas) result.
+    assert res.name == "sum_ab"  # pandas series has a name and we honor it
+    # Rename res, just for `assert_frame_equal`.
+    assert_frame_equal(res.rename("data"), cudf.Series([3.0, 5.0]))
```
gforsyth, line 38: Does legate guarantee result order?  For something this small it's unlikely to drift, but if ordering isn't broadly guaranteed, it's good to have that in mind when writing tests (and adding explicit `order_by`s to expected results)

python/tests/test_ibis.py
```diff
@@ -0,0 +1,136 @@
...
+            """\
+        a,b,c
+        1,2,3
+        2,3,4
+        """
+        )
+    )
+
+    schema = ibis.schema({"a": "int64", "b": "float64", "c": "int64"})
+    b = LegateBackend().connect()
```
gforsyth, line 28:
```suggestion
    b = LegateBackend.connect()
```

The `connect` call will return a new class object anyway, so no need to instantiate the backend class separately.

gforsyth, line 28: by convention, we use `con` as the connection object

gforsyth (commented):
Hey @seberg, this looks like a really great start!  I probably need more familiarity with Legate Dataframe to do a properly thorough review, but there's nothing here that stands out as "trouble".  

I'm glad this came together relatively easily.  

I left a few comments inline, mostly trying to answer questions.

python/tests/test_ibis.py
seberg: Yes, order is stable legate (and except for some manual operations, that is even pretty deep if we use the legate data model).

python/legate_dataframe/experimental_ibis/schema.py
seberg: Ah, logically, all columns are nullable.  Sometimes we may know that there are nulls as an optimization.
So, I think that means that we can reject it here (it won't arise without user input as an optimization hint).  At least for now.

python/legate_dataframe/experimental_ibis/__init__.py
seberg: OK, I read that as this check making sense and will remove the comment, easy to soften the error eventually (I am not sure the pandas/polars backend were/are strict about it, so the question).

python/legate_dataframe/experimental_ibis/__init__.py
seberg: Sounds like this is fine, and keeping `NotImplementedError` also where it is used  (will just remove the comment for now).

(Specific errors, as long as there are not too many, does seem nice.)

python/legate_dataframe/experimental_ibis/__init__.py
seberg: Ah OK, somehow I thought I saw an error for it.  I guess that means just ignoring `limit` for now, thanks!

seberg: For reference, I guess I got it from here: https://github.com/ibis-project/ibis/blob/70de8db00b83dd3652a0e1e123711ee80b68983e/ibis/backends/pandas/__init__.py#L314

Maybe it would make sense to at least support it by slicing the final (full) result.
(Although, the nice way would be to implement a/the `ops.Limit` which then could be optimized to be pushed down as far as possible as a re-write.

python/tests/test_ibis.py
seberg: Hmmm, realized I hadn't applied this, and just tried it.  But it seems `connect()` is a normal method and the instantiation is necessary?

PR #20, opened 2025-01-30
ENH: Add ``apply_boolean_mask``

seberg:
Adding it with a similar file structure to libcudf which uses
``stream_compaction``.  (Not sure about the names, but also good to
keep aligned with libcudf.)

Straight forward exposure of the table-wise boolean masking.

---

Other fixup is due to sorting tests running into issues with the high column things now, so lowered it.
(It must be unrelated, but the `test_task.py` seemed to hang when run as an individual test.)

### DIFF
--- a/python/legate_dataframe/lib/stream_compaction.pyx
+++ b/python/legate_dataframe/lib/stream_compaction.pyx
@@ -38,7 +38,6 @@ def apply_boolean_mask(
     Returns
     -------
         The ``LogicalTable`` containing only the rows where the boolean_mask was true.
-
     """
     return LogicalTable.from_handle(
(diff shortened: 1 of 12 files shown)


python/legate_dataframe/lib/stream_compaction.pyx
```diff
@@ -0,0 +1,44 @@
...
+    ----------
+    tbl
+        The table to filter.
+    boolean_mask
+        The boolean mask to apply.
+
+    Returns
+    -------
+        The ``LogicalTable`` containing only the rows where the boolean_mask was true.
+
```
madsbk, line 41:
```suggestion
```

madsbk (approved):
Looks good!

PR #21, opened 2025-02-04
BUG: Sort code must deal with zero row local chunks

seberg:
The local chunks may be empty and the sort code must make sure to deal with it:
* An empty local table cannot share split points (the way it is done here)
* As no split points are shared, the total number of split points may be fewer than the number of ranks.

We need to deal with both of these.  Now that there is boolean masking it can be used to trigger these paths reliably (currently at least).

### DIFF
--- a/cpp/src/sort.cpp
+++ b/cpp/src/sort.cpp
@@ -60,16 +60,21 @@ std::unique_ptr<cudf::column> get_split_ind(GPUTaskContext& ctx,
                                             int nsplits,
                                             bool include_start)
 {
-  auto nvalues_per_split = nvalues / nsplits;
-  auto nvalues_left      = nvalues - nvalues_per_split * nvalues;
-  if (nvalues_per_split == 0) {
-    nsplits = nvalues_left;  // Only return non-empty splits
+  cudf::size_type nvalues_per_split, nvalues_left;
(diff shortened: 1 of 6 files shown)


cpp/src/sort.cpp
```diff
@@ -300,28 +319,32 @@ class SortTask : public Task<SortTask, OpCode::Sort> {
...
-
     // Create a new locally sorted table (we always need this)
     auto cudf_tbl  = tbl.table_view();
     auto key       = cudf_tbl.select(keys_idx);
     auto sort_func = stable ? cudf::stable_sort_by_key : cudf::sort_by_key;
     auto my_sorted_tbl =
       sort_func(cudf_tbl, key, column_order, null_precedence, ctx.stream(), ctx.mr());
 
-    if (ctx.nranks == 1 || tbl.is_broadcasted()) {
+    if (ctx.nranks == 1) {
```
seberg, line 329: The broadcast check needs fixing, it isn't reliable enough.  I think here the `nrank` should be sufficient, though.  (in joins this is probably a latent problem)

madsbk (approved):
LGTM

seberg:
Thanks, had asked Wonchan about the barriers and that seems fine. (I wonder if it is overly defensive but...).

PR #22, opened 2025-02-06
BLD: Split package build to not include tests and prep for 25.01

seberg:
This ensures that we pick up the 25.01 release to cut an accompanying legate-df 25.01 release.
(I'll note that as of now, using the nightly is probably a good idea in general.)


seberg:
I bloated this PR (a lot), and implemented proper package splitting.

cpp/cmake/thirdparty/get_gtest.cmake
```diff
@@ -17,25 +17,7 @@ function(find_and_configure_gtest)
...
-      GLOBAL_TARGETS gtest gmock gtest_main gmock_main
-      NAMESPACE GTest::
-    )
-
-    include("${rapids-cmake-dir}/export/find_package_root.cmake")
-    rapids_export_find_package_root(
-      BUILD GTest [=[${CMAKE_CURRENT_LIST_DIR}]=] EXPORT_SET LegateDataframe-testing-exports
-    )
-  endif()
+  rapids_cpm_gtest(BUILD_STATIC)
```
seberg, line 20: Both aligns with `libcudf` and should also help that we don't install gtest libraries by accident (or are missing them if we don't install them).

cpp/tests/test_repartition_by_hash.cpp
```diff
@@ -38,8 +38,9 @@ namespace {
 static const char* library_name = "test.repartition_by_hash";
 
 struct CheckHash : public legate::LegateTask<CheckHash> {
-  static constexpr auto TASK_ID             = legate::LocalTaskID{0};
-  static constexpr auto GPU_VARIANT_OPTIONS = legate::VariantOptions{}.with_has_allocations(true);
+  static constexpr auto TASK_ID = legate::LocalTaskID{0};
+  static constexpr auto GPU_VARIANT_OPTIONS =
+    legate::VariantOptions{}.with_has_allocations(true).with_concurrent();
```
seberg, line 43: I guess this was a bit random, this should have failed on main already...

.github/workflows/conda-python-build.yaml
```diff
@@ -18,7 +18,7 @@ env:
   CUDAARCHS: "all-major"
   # where conda-python-build puts files it creates
   RAPIDS_CONDA_BLD_OUTPUT_DIR: /tmp/conda-bld-output
-
+  GH_TOKEN: ${{ github.token }}
```
seberg, line 21: I had to add this for the sake of `source rapids-date-string`, but it seems to me that this works differently in `cudf` (maybe via `date: ${{ inputs.date }}`?! but if I read `rapids-date-string` the only thing that should work is `GITHUB_ACTIONS!=true`)

conda/recipes/legate-dataframe/meta.yaml
```diff
@@ -70,32 +69,60 @@ requirements:
...
+  - name: legate-dataframe
+    version: {{ version }}
+    string: "cuda{{ cuda_major }}_py{{ py_version }}_{{ PKG_BUILDNUM }}"
+    script: install_legate-dataframe.sh
+    requirements:
+      build:
+        - cmake {{ cmake_version }}
+      host:
+        - cuda-version ={{ cuda_version }}
+        - python
```
seberg, line 84: Without Python here, I think I got a Python 3.13 rather than the same Python (at in my local build).

jameslamb: `cuda-version` in `host` will help ensure you get the expected version of `nvcc` in the `host` environment, I think. This is good to keep.

.github/workflows/conda-python-build.yaml
jameslamb: Ahhhh yeah this is probably necessary because `rapids-data-string` uses the GitHub CLI like this:

```shell
WORKFLOW_DATE=$(rapids-retry gh run view "${GITHUB_RUN_ID}" --json createdAt | jq -r '.createdAt')
```

([build link](https://github.com/rapidsai/gha-tools/blob/30a8ac5fcce9284ea8443c636999ab49e5c563ea/tools/rapids-date-string#L6))

`cudf` and other RAPIDS repos get that secret automatically because they use the workflows from the `shared-workflows` repo, which sets this env variable:

https://github.com/rapidsai/shared-workflows/blob/dc61137d6619d4131586915f41251b84a3bbcdf4/.github/workflows/conda-python-build.yaml#L124-L125

And they use `secrets: inherit`.

```yaml
  conda-python-build:
    needs: conda-cpp-build
    secrets: inherit
    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@nvks-runners
```

https://github.com/rapidsai/cudf/blob/6dc4a536897b3156a3b549a400113e0373798dc9/.github/workflows/pr.yaml#L158-L161

We're not getting those inherited secrets here because we're not using the workflows from `rapidsai/shared-workflows`. That's the same reason that `sccache` setup in these `rapidsai/legate-*` requires additional work you don't see across the other RAPIDS repos (ref: https://github.com/rapidsai/legate-raft/issues/2).

It'd be awkward to start using the `rapidsai/shared-workflows` workflows here, because they bake in:

* different support matrix of Python version, CUDA version, operating system
* different calendar versioning scheme (24.12, 25.02, 25.04, etc.) than Legate (24.11, 25.01, 25.03)

jameslamb: So all that to say... yes this is necessary. Can you move it up to keep this `env:` list alphabetized, and add a comment like this?

```yaml
  # needed for some uses of rapidsai/gha-tools in build scripts, which
  # use the GitHub CLI
  GH_TOKEN: ${{ github.token }}
```

jameslamb (approved):
The recipe split looks good to me! I left a few small suggestions... none are blocking, and I don't need to re-review after you've addressed them.

I support doing this and following a plan similar to https://github.com/rapidsai/legate-boost/pull/208#pullrequestreview-2602106225 to fix the dependencies in a `25.01.01` of `legate-dataframe`.

PR #23, opened 2025-02-07
ENH: Implement column selection for parquet reading

seberg:
Yes, eventually we should add more options (and maybe use the opts), but columns selection is fairly important and at least on the Python side, maybe an iterable is a decent longer solution.


madsbk (approved):
looks good

PR #25, opened 2025-02-10
BLD: Ensure we fetch the `_gpu` tag version

seberg:
I am not sure if this should also be included in the dependencies.yaml, but I suppose it shouldn't matter too much there for a release at least. (conda didn't seem to like just adding it?)

This is identical to legate-boost (which needs to pick cpu or gpu depending on the build).

### DIFF
--- a/conda/recipes/legate-dataframe/meta.yaml
+++ b/conda/recipes/legate-dataframe/meta.yaml
@@ -55,7 +55,7 @@ requirements:
     # the '<=12.6.0' can be removed once this is resolved:
     # https://github.com/conda-forge/cuda-python-feedstock/issues/95
     - cuda-python >=12.0,<13.0a0,<=12.6.0
-    - legate {{ legate_version }}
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
> depencies.yaml ... conda didn't seem to like just adding it?

Not sure what specifically you saw, but it should be fine to include this for `conda`-only dependency lists in `dependencies.yaml`. Be sure not to have a leading space in front of the first `*`, so it isn't interpreted as a YAML anchor.

But it also should not be necessary... `dependencies.yaml` is not involved in updating this conda recipe, and if you're using the conda environments it creates on a system with a GPU, you should automatically end up with `*_gpu*` builds of `legate` via `legate` GPU builds' runtime dependency on the `__cuda` virtual package.

seberg:
Sorry, should have linked the build, I had tagged 25.01.00 and the build failed: https://github.com/rapidsai/legate-dataframe/actions/runs/13236972318

PR #26, opened 2025-02-10
BLD: Fix split recipe dependencies [skip ci]

seberg:
Run-exports is only picked up from the current build step (which makes sense, since they can differ), but this means that we need to manually include anything that we want to pick up in the host/build section (i.e. mostly legate).

Since cudf/rmm already depend on pylibcudf, libcudf, and librmm, I have not duplicated these.

**Do not merge until version upload is clarified.**

### DIFF
--- a/conda/recipes/legate-dataframe/meta.yaml
+++ b/conda/recipes/legate-dataframe/meta.yaml
@@ -72,10 +72,14 @@ outputs:
     string: "cuda{{ cuda_major }}_py{{ py_version }}_{{ PKG_BUILDNUM }}"
     script: install_legate-dataframe.sh
     requirements:
-      # Note: cuda-version and legate (+cupynumeric) are here for their run-exports
+      # Note: build/host exists for run-exports (except cmake, pip, and python)
+      # for librmm, libcudf, pylibcudf we rely on the rmm/cudf run dependency.
       build:
         - cmake {{ cmake_version }}
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
         - cuda-version ={{ cuda_version }}
+        - {{ stdlib("c") }}
       host:
         - cuda-version ={{ cuda_version }}
         - legate {{ legate_version }} =*_gpu*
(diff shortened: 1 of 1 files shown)


seberg:
OK, so the artifacts here contain:
```
    "__cuda",
    "cuda-cudart",
    "cuda-version >=12,<13.0a0",
    "cudf 24.12.*",
    "legate 25.01.00 *_gpu*",
    "numpy >=1.23,<3.0.0a0",
    "python >=3.10,<3.11.0a0",
    "python_abi 3.10.* *_cp310",
    "rmm 24.12.*"
```
while the current upload incorrectly contain:
```
__cuda, cuda-cudart, cuda-version >=12,<13.0a0, cudf 24.12.*, legate, numpy >=1.23,<3.0.0a0, python >=3.12,<3.13.0a0, rmm 24.12.*
```
and the pre-split uploads included:
```
__cuda, __glibc >=2.17,<3.0.a0, cuda-cudart, cuda-version >=12,<13.0a0, cudf 24.12.*, legate 25.01.00.dev124 *_gpu*, libcudf >=24.12.0,<24.13.0a0, libgcc, libgcc-ng >=12, librmm >=24.12.1,<24.13.0a0, libstdcxx, libstdcxx-ng >=12, numpy >=1.23,<3.0.0a0, python >=3.11,<3.12.0a0, rmm 24.12.*
```

@jameslamb does the seem right to you?  Then I can put this in (maybe with a CI skip) and tag another release...  Although, in this case in theory it might be nice to actually delete the old releases, because they'll incorrectly match far too much :/.

conda/recipes/legate-dataframe/meta.yaml
```diff
@@ -98,6 +97,8 @@ outputs:
         - numpy >=1.23,<3.0.0a0
         - rmm {{ rapids_version }}
         - cudf {{ rapids_version }}
+    ignore_run_exports_from:
+      - cupynumeric  # only a build dependency to help resolver
     run_exports:
       - {{ pin_subpackage("legate-dataframe", max_pin="x.x") }}
```
seberg, line 103: I think I may have changed this, `legate-raft` now includes `legate-dataframe` twice, one with: `legate-dataframe >=25.1.0,<25.2.0a0`.  I guess that seems correct?  (although unnecessary, and I could either ignore the run-export or omit the manual `legate-dataframe =25.01.*,>=0.0.0.dev0` there)

seberg:
Since I am pretty confident about librmm/libcudf/pylibcudf didn't add those, but just add the compilers back (that weren't ignored anyway), so hopefully that is safe.
(Sorry, I really don't have a feeling for best practices here.)

PR #29, opened 2025-02-18
API: Expose legates ``offload_to`` method to columns and tables

seberg:
Since the name is ``offload_to`` I didn't really like putting a default (felt `col.offload_to()` reads weird).

So like legate, the argument must be passed for now even if it is always expected to be SYSMEM, at least for the moment.

The test works, but might fail if the config is not set.  Plus in many combination it will just skip the test.  The test may also be slow (on the default they take together about 10s in local tries, so I assume it is OK for now.)

But... if they make problems, we should just disable them...

---

Conftest wasn't used anymore, I am always running via `legate` now.  Maybe it works again, it didn't for a while.  But for now I think `./ci/run_pytests.py` is OK.
Some test modifications for local stability (adjusted for smaller default memory and the many rows test somehow created problems now, tracking that as a potential legate issue)


RAMitchell (approved):
Seems generally useful to have! I'd like to avoid the user having to do this but this could be a short term solution or a building block to something more user friendly.

seberg:
Thanks for the review.  Will put this in as it may be useful for benchmarks/enabling some experiments.

(Happy to follow up eventually, I am not quite sure this is the best API, but it is the same as for `LogicalArray`.)

PR #32, opened 2025-02-24
ENH: Implement basic reductions

seberg:
This implements basic aggregations using a similar pattern to the simple aggs/finalizer used in parts of cudf.

Since that pattern doesn't take into account the initial value and is in the `detail` API, opted instead to create a custom helper.

That is, we use `(dtype, agg, initial)` to identify each reduce and then split the reduce into two steps:
1. A first-pass that reduces only locally.
2. A finalize pass that works with local results or other final
   results.

Supporting further aggregations should be straight forward additions in the future.

The design should generalize relatively well to re-use for other aggregations (e.g. maybe to improve groupby-aggs) or to calculate multipe aggregations for the same column (since e.g. many need the sum or count temporary result).

The one thing that should be improved is that "count valid" is relatively special and it should possibly become a specialized task and be exposed as explicit API (with `count_nulls()`, we actually should have both probably because `num_rows()` is blocking).

---

I'll note that I did see occasional hangs depending on how tests are run (multi-gpu).  I suspect they are identical to earlier hangs I have seen that seem to originate in legate/legion.

### DIFF
--- a/cpp/src/reduction.cpp
+++ b/cpp/src/reduction.cpp
@@ -118,9 +118,7 @@ class ReduceLocalTask : public Task<ReduceLocalTask, OpCode::ReduceLocal> {
 
     // Note: cudf has no helper to go to a column view right now, but we could
     // specialize this in principle.
-    auto ret = cudf::make_column_from_scalar(*scalar_res, 1, ctx.stream(), ctx.mr());
-
-    output.move_into(std::move(ret));
+    output.move_into(cudf::make_column_from_scalar(*scalar_res, 1, ctx.stream(), ctx.mr()));
   }
 };
 
@@ -386,7 +384,6 @@ class AggregationHelper final {
       }
       default: {
         auto first_pass_res = get_first_pass_result(agg, output_type);
-        // TODO: Use initial here (we may need it, e.g. for sums).
         return perform_simple_reduce(
           first_pass_res, agg.kind, /* finalize */ true, output_type, initial);
       }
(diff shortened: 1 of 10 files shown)


cpp/src/reduction.cpp
```diff
@@ -0,0 +1,433 @@
...
+        auto sum        = get_result(*cudf::make_sum_aggregation(), output_type, std::nullopt);
+        auto cudf_int64 = cudf::data_type{cudf::type_id::INT64};
+        auto counts     = get_result(*cudf::make_count_aggregation(), cudf_int64, std::nullopt);
+
+        return legate::dataframe::binary_operation(
+          sum, counts, cudf::binary_operator::DIV, output_type);
+      }
+      default: {
+        auto first_pass_res = get_first_pass_result(agg, output_type);
+        // TODO: Use initial here (we may need it, e.g. for sums).
```
seberg, line 389:
```suggestion
```

cpp/src/reduction.cpp
```diff
@@ -0,0 +1,433 @@
...
+      } else {
+        scalar_res = cudf::reduce(col_view, *agg, output.cudf_type(), ctx.stream(), ctx.mr());
+      }
+    }
+
+    // Note: cudf has no helper to go to a column view right now, but we could
+    // specialize this in principle.
+    auto ret = cudf::make_column_from_scalar(*scalar_res, 1, ctx.stream(), ctx.mr());
+
+    output.move_into(std::move(ret));
```
madsbk, line 123:
```suggestion
    output.move_into(cudf::make_column_from_scalar(*scalar_res, 1, ctx.stream(), ctx.mr()));
```

madsbk (approved):
Looks good to me

PR #33, opened 2025-02-25
API,BUG: Remove `global_num_rows`, use `is_partitioned`

seberg:
This is breaking as we stop passing around `global_num_rows`, but this
should improve legate's ability to parallelize task launches (I think)
since fetching the `global_num_rows` on an unbound table requires blocking.

`is_partitioned` should also be a safer API here, because the local and
global number of rows could be equal if all but one chunks have 0 entries
(which is may be hard to trigger but is possible with filtering).

---

* I removed `global_num_rows` for now.  It may well be useful again eventually.
* The test did hang without the fix, but not 100% sure if that was actually the bug I expected :).


madsbk (approved):
Looks good, nice to get rid of `global_num_rows`!

PR #36, opened 2025-02-27
API: Introduce scalar columns

seberg:
We don't want to deal with specialized scalar objects, but scalars are special at least in that they are always length 1.

We could always allow broadcasting length 1 columns (allow them as scalar). However, checking the number of rows is blocking if we use unbound stores (which we normally do).
This is undesireable as task launch (e.g. for binary ops) may need to be different (broadcast scalars).

General notes:
* `replace_nulls` usually doesn't need to return a nullable result but without blocking on the scalar we don't know that for sure. (Not a regression, but it may be nice to make an exception here.)
* I decided not to propagate "scalar" in e.g. `empty_like` by default. It seemed simpler to err on the "column result" side and lose the scalar information unless explicitly preserved.
* This does enable "scalars" in unary/binary ops.
* It means we implicitly allow "scalars" instead of columns by default. That seems mostly fine to me (one could reject explicitly if needed).

Note that this PR does not optimize scalar handling in any way.  It may make sense to specialize unpacking/packing the scalar in the PhysicalColumn.


python/legate_dataframe/lib/core/scalar.pyx
```diff
@@ -1,42 +1,49 @@
...
+from legate_dataframe.lib.core.column cimport LogicalColumn
+
+ScalarLike = (
+    numpy.number | numbers.Number | cudf.Scalar | legate.core.Scalar | DeviceScalar
+)
 
 
-cdef cpp_ScalarArg cpp_scalar_arg_from_python(scalar: ScalarLike):
-    """Convert any supported Python scalar to a scalar task argument
+cdef LogicalColumn cpp_scalar_col_from_python(scalar: ScalarLike):
```
seberg, line 22: This is very forgiving to allow conversion from all of these...

I suspect we should just disallow it eventually.  I also still relies on the non `pylibcudf` version for Python facing things (I am not sure if I can transition on cudf 24.12).

(That probably also means changing what `to_cudf()` and `to_cudf_scalar()` return, which is breaking, but seems OK.  Changing what they accept is a bit more churn.)

cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -376,6 +477,7 @@ class PhysicalColumn {
   const cudf::data_type cudf_type_;
   mutable std::vector<std::unique_ptr<cudf::column>> tmp_cols_;
   mutable std::vector<rmm::device_buffer> tmp_null_masks_;
+  const bool scalar_out_;  // scalar output checks binding has size 1.
```
seberg, line 480: Opted to call this differently, because it seemed to matter less here and I ended up only using it to check if binding a result adds up.

cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -200,17 +250,33 @@ class LogicalColumn {
...
+  /**
+   * @brief Return true if the column is considered scalar.
+   *
+   * A scalar column always has a size of one.  Scalar columns are mostly supported
+   * for binary operations and have scalar/broadcast semantics if supported.
+   * (For unsupported operations they may work, but behave like a length 1 column).
+   *
+   * @return true if scalar, otherwise false.
+   */
+  bool scalar() const { return scalar_; };
```
madsbk, line 278: nit: maybe rename to `is_scalar`, to make it clear that it doesn't return a _scalar_ ?

seberg: Thanks for the nudge, done.  Had wondered briefly but not acted up on it :).

madsbk (approved):

seberg:
Thanks for the review.  Working on aggs, I noticed that one assert was wrong and another check didn't work in practice, so pushed a tiny fix for those and will merge once tests pass.

PR #37, opened 2025-03-06
API: Add `to_array()` methods to convert column or tables to arrays

seberg:
I opted for `to_array()` because `to_cupynumeric()` just seems very long, but happy to change!
(But happy to change that, I am not sure what I think is better.)

It may be that, especially for columns, `cupynumeric.asarray(col)` should just be the long-term solution.

For tables, I am personally not sure:  `pandas` does it, but in practice implicitly converting columns into a new dimension isn't that obvious to me.  Tables and arrays are very different.

(Technically breaking, because previously converting times was somewhat sensible, and now it is reject unless you use `get_logical_array()` explicitly.)


python/legate_dataframe/lib/core/column.pyx
```diff
@@ -178,6 +198,52 @@ cdef class LogicalColumn:
...
+        ValueError
+            If the column has a mask and there are masked values.
+        """
+        # Delay import of cupynumeric.
+        from cupynumeric import asarray
+
+        array = self.get_logical_array(check_dtype=True)
+        if array.num_children != 0:
+            raise TypeError(
+                "cupynumeric doesn't support arrays with children (e.g. strings)."
```
seberg, line 232: Cupynumeric went into a long exception here (not sure it was critical).  So opting to just deal with it explicitly...

seberg:
@danielfrg does this look good to you to simplify conversion to cupynumeric?

danielfrg:
Yes, that looks exactly what we need.

I do like `.to_array()` more than `.to_cupynumeric()`.

python/legate_dataframe/lib/core/column.pyx
```diff
@@ -144,19 +147,36 @@ cdef class LogicalColumn:
...
         return self._handle.is_scalar()
 
-    def get_logical_array(self) -> LogicalArray:
+    def get_logical_array(self, *, check_dtype=False) -> LogicalArray:
         """Return the underlying logical array
 
+        Parameters
+        ----------
+        check_dtype
+            If ``True`` check that the dtype round-trips.  Default's to False.
```
madsbk, line 156:
```suggestion
            If ``True`` check that the dtype round-trips.  Defaults to False.
```

madsbk (approved):
Looks good!

PR #41, opened 2025-03-10
MAINT: Reduce the number of scalars passed to tasks

seberg:
This packs scalars into vectors.  Legate (now?) supports `std::vectors<type>` just fine for the most part, but since it unpacks them as a `span` (which is probably good), opted to keep the old pattern for now.

Unfortunatley, `bool` and `string` vectors are not supported directly so this uses overloads.

I did this hoping to resolve some issues that, in the end, didn't resolve.  But while it is a bit tedious with the table, it seems like a step in the right direction anyway.
(We should probably use the `span`, vector as scalar directly in some places, and teach `legate` to deal with bool and string vectors.)


cpp/include/legate_dataframe/core/task_argument.hpp
```diff
@@ -61,12 +61,41 @@ inline void add_next_scalar<legate::Scalar>(legate::AutoTask& task, const legate
  * @param items The vector of scalars.
  */
 template <typename T>
-void add_next_scalar_vector(AutoTask& task, const std::vector<T>& scalars)
+inline void add_next_scalar_vector(AutoTask& task, const std::vector<T>& scalars)
```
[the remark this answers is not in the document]
madsbk: I don't think you need inline as this is a template?

cpp/include/legate_dataframe/core/task_argument.hpp
```diff
@@ -63,10 +63,39 @@ inline void add_next_scalar<legate::Scalar>(legate::AutoTask& task, const legate
...
-    add_next_scalar(task, scalar);
+  // Temporary work-around 2025-03.  Legate may assert on data even for 0-size
+  // so we build the scalar manually and do not copy if it is zero size.
+  auto arr_type =
+    legate::fixed_array_type(legate::primitive_type(legate::type_code_of_v<T>), scalars.size());
+  add_next_scalar(task, legate::Scalar(arr_type, scalars.data(), /* copy */ scalars.size() > 0));
+}
+
+template <>
+inline void add_next_scalar_vector(AutoTask& task, const std::vector<bool>& scalars)
```
madsbk, line 74:
```suggestion
void add_next_scalar_vector(AutoTask& task, const std::vector<bool>& scalars)
```

cpp/include/legate_dataframe/core/task_argument.hpp
```diff
@@ -63,10 +63,39 @@ inline void add_next_scalar<legate::Scalar>(legate::AutoTask& task, const legate
...
+  // Also do not copy if empty (see general vector code above).
+  auto arr_type = legate::fixed_array_type(legate::bool_(), scalars.size());
+  std::vector<char> tmp;
+  tmp.assign(scalars.begin(), scalars.end());
+
+  add_next_scalar(task, legate::Scalar(arr_type, tmp.data(), /* copy */ scalars.size() > 0));
+}
+
+template <>
+inline void add_next_scalar_vector(AutoTask& task, const std::vector<std::string>& scalars)
```
madsbk, line 87:
```suggestion
void add_next_scalar_vector(AutoTask& task, const std::vector<std::string>& scalars)
```

cpp/include/legate_dataframe/core/table.hpp
```diff
@@ -471,23 +471,29 @@ std::vector<legate::Variable> add_next_output(legate::AutoTask& task, const Logi
 template <>
 inline task::PhysicalTable get_next_input<task::PhysicalTable>(GPUTaskContext& ctx)
```
madsbk, line 472:
```suggestion
task::PhysicalTable get_next_input<task::PhysicalTable>(GPUTaskContext& ctx)
```

cpp/include/legate_dataframe/core/table.hpp
```diff
@@ -471,23 +471,29 @@ std::vector<legate::Variable> add_next_output(legate::AutoTask& task, const Logi
...
+  cols.reserve(cudf_type_ids.size());
+  for (auto& cudf_type_id : cudf_type_ids) {
+    auto cudf_type = cudf::data_type{static_cast<cudf::type_id>(cudf_type_id)};
+    cols.push_back(task::PhysicalColumn(ctx, ctx.get_next_input_arg(), cudf_type));
   }
   return task::PhysicalTable(std::move(cols));
 }
 
 template <>
 inline task::PhysicalTable get_next_output<task::PhysicalTable>(GPUTaskContext& ctx)
```
madsbk, line 486:
```suggestion
task::PhysicalTable get_next_output<task::PhysicalTable>(GPUTaskContext& ctx)
```

madsbk (approved):
Looks good, but think there is more `inline` we can remove

PR #42, opened 2025-03-10
TST: Clean up temporary cpp test directories

seberg:
This was probably useful in debugging at some point, but survived and got cargo-culted.  I noticed that the tmp folder gets filled up a lot (I think to the point of tests not passing).


cpp/tests/test_csv.cpp
```diff
@@ -37,7 +37,7 @@ TYPED_TEST_SUITE(NumericCSVTest, cudf::test::NumericTypes);
 
 TYPED_TEST(NumericCSVTest, ReadWrite)
 {
-  TempDir tmp_dir{false};
+  TempDir tmp_dir;
```
seberg, line 40: `false` here means no cleanup, the default is to cleanup.

PR #46, opened 2025-03-21
Update to legate 25.05 and align task config

seberg:
This lists all the tasks explicitly now, the registrar pattern is probably salvangable but I am not sure that legate wants to. (At which point we could vendor it, but maybe this isn't so bad.)

---

We probably should just put this in soon...  Although, there is a very good chance that we either can revert most of the changes again or need to restructure them differently due to deprecations.


madsbk (commented):
Would it make sense to have a helper template function that returns `legate::AutoTask` given a `Task` class?

madsbk (approved):
OK, let's wait for the Legate team to settle on a design :)

seberg:
I guess I'll put it in and hope I can revert/make it nicer in the future (it would be nice to have up-to date builds).

By now changes in main should have made the other PR a bit nicer as well, probably...

PR #48, opened 2025-04-03
API: Allow parquet reading into (transposed) array

seberg:
This adds a call to do a chunked copy directly into a transposed array. To be honest, the only reason to do this inside legate-df, is that we have the cudf dependency here anyway.
To some degree, the reason is just to work-around current legate issues with too many columns.

---

Always trickier then I expect... @RAMitchell this should allow reading in these huge tables, effectively directly to cupynumeric via `nullable=False` and then just using `cupynumeric.asarray()`.

I am not sure if the chunking choice here is ideal, i.e. whether this is as faster as it can be.  The tranpose-copy is very similar to the "interleaved copy" in cudf, and I can imagine that it could be sped up as well.

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2024, NVIDIA CORPORATION.
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
 
 set -e -E -u -o pipefail
 
@@ -44,4 +44,6 @@ rapids-mamba-retry install \
 
 rapids-print-env
 
+# Make documentation, we need gpu but otherwise minimal resources
+LEGATE_CONFIG="--omps=0 --gpus=1 --fbmem=100 --cpus=1 --utility=1" \
 make -C docs html
--- a/ci/run_ctests.sh
+++ b/ci/run_ctests.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2024, NVIDIA CORPORATION.
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
 
 set -euo pipefail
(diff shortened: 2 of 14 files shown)


RAMitchell (commented):
You may want to follow the pandas api in terms of missing values and type coercion:
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_numpy.html

RAMitchell:
Would be good if we can make some performance comparisons in terms of both time and peak memory in this thread, to confirm that we have unblocked this issue.

seberg:
This mainly solves the memory issue and yeah I can load the 10_000_000x5000 file with this, I am not sure about the performance overhead of the copy yet.

A `null_value=` and `dtype=` API make sense to me, although I don't think I want to do auto-promotion here for now.

RAMitchell (approved):

seberg:
I changed the API to `null_value=` and `type=`.  That is indeed better and more useful.

However, at least for now, it'll be a bit less convenient because I am asking you to use a `legate.Scalar()` for the `null_value` and make sure it's type matches the result type (things that can be improved, though).
So when loading data that shouldn't contain any nulls _and_ you want to convert that to cupynumeric, you need to pass the right `legate.core.Scalar()`

I will be OoO, for a bit soon, so if this approach makes sense to you, it may make sense to put this.  I agree, we should double check performance, I also didn't tune the chunked reader size (I am not aware sure it should matter a lot, but maybe there is a sweet spot).

(So yes, should keep the issue open probably to at least check this.)

seberg:
Unclear why CI isn't running, will try close/reopen...

seberg:
OK, tests passing now with some adjustment for legate launching.  I'll put this in and that'll also make it easier to test it once more.

PR #49, opened 2025-04-25
API: Add `unaryop.cast` to cast a single column

seberg:
Add API to allow casting a single column (for now functional).

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,14 @@ This is **not** a drop-in replacement of [Pandas](https://pandas.pydata.org/), i
 
 In the future, we plan to introduce a high-level pure Python package that implements all the nice-to-have features known from Pandas using the low-level API's primitives.
 
+[Python API and further documentation](https://rapidsai.github.io/legate-dataframe/).
+
 ## Install
 
 You can install `legate-dataframe` packages from the [conda legate channel](https://anaconda.org/legate/)
 using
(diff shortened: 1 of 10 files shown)


seberg:
One reason for this, is that we wanted a `datetime -> numeric` path/cast.  Unfortunately, that isn't actually supported, however `timedelta -> numeric` is.

That means that the following seems to work:
```
diff = binary_operation(series, cudf.Scalar("1971-01-01", dtype="M8[s]"), binary_operator.SUB, "m8[s]")
cast(diff, "float64")
```
(the unit `"s"` may have to match precisely with that of the series and unfortunately this is not reliably caught outside of the task right now.)

Some casts also fail hard, it seems that [`is_supported_cast`](https://docs.rapids.ai/api/libcudf/stable/group__transformation__unaryops.html#ga773a19e1a898ceecb22eb05f6a6b7a78) isn't fully reliable.

RAMitchell (approved):
LGTM. Should we give feedback here to the cudf team?

seberg:
I'll let them know.  I think it is maybe an issue, but a very small one.  `libcudf` uses specialized from/to string functions for string conversion.
Here if we integrate it with `cast` that goes away, if we would also extend it with specialized functions eventually, then it might make sense to reject non numeric, etc. columns explicitly.

PR #51, opened 2025-05-05
Let legate allocate stores

### LINKED ISSUE #50: Google tests fail on multi-gpu
```
CUDA_VISIBLE_DEVICES=0,1 ./cpp/build/gtests/cpp_tests --gtest_filter=*NumericCSVTest/1*
Note: Google Test filter = *NumericCSVTest/1*
[==========] Running 10 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 5 tests from NumericCSVTest/1, where TypeParam = short
[ RUN      ] NumericCSVTest/1.ReadWrite
/home/nfs/rorym/anaconda3/envs/legate/src/cudftestutil/utilities/column_utilities.cu:562: Failure
Failed
first difference: lhs[1] = 0, rhs[1] = 1
Google Test trace:
/home/nfs/rorym/legate-dataframe/cpp/tests/test_csv.cpp:56:  <--  line of failure

/home/nfs/rorym/anaconda3/envs/legate/src/cudftestutil/utilities/column_utilities.cu:562: Failure
Failed
first difference: lhs[0] = 0, rhs[0] = 4
Google Test trace:
/home/nfs/rorym/legate-dataframe/cpp/tests/test_csv.cpp:56:  <--  line of failure

cpp_tests: /tmp/conda-croot/legate/work/arch-conda/skbuild_core/_deps/legion-src/runtime/realm/mem_impl.inl:648: void Realm::BasicRangeAllocator<RT, TT>::deallocate(TT, bool) [with RT = long unsigned int; TT = Realm::RegionInstance]: Assertion `missing_ok' failed.``
```

RAMitchell:
## Description
Changes the way we allocate a legate store and copy an external buffer into it.
Closes #50

### DIFF
--- a/python/tests/test_table.py
+++ b/python/tests/test_table.py
@@ -10,6 +10,7 @@ from legate_dataframe.lib.stream_compaction import apply_boolean_mask
 from legate_dataframe.testing import assert_frame_equal, guess_available_mem
 
 
+@pytest.mark.skip(reason="This causes CI hangs. Investigate rewriting this test.")
 def test_offload_to():
     # Note that, if `LEGATE_CONFIG` is set but not used, this may currently fail.
     available_mem_gpu, available_mem_cpu = guess_available_mem()
--- a/cpp/src/core/column.cu
+++ b/cpp/src/core/column.cu
@@ -53,17 +53,11 @@ namespace {
   legate::Runtime* runtime = legate::Runtime::get_runtime();
 
   if (nelem == 0) { return runtime->create_store({nelem}, dtype); }
-
-  // For now, we copy the device buffer to host memory and hand it over to legate with the
-  // `std::free()` function.
-  // TODO: avoid this copy: <https://github.com/rapidsai/legate-dataframe/issues/80>
   const size_t nbytes = nelem * dtype.size();
-  void* host_buffer   = std::malloc(nbytes);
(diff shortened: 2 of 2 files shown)


seberg:
Very strange failure in CI.... The Task is claiming that the file doesn't exist, but we pass in the file after having already checked its existence.  Unfortunately, I didn't see the issue locally (although here it only happened with the Python 3.12 test run).

Pushed a silly commit to see what those files are, not sure it'll help...

RAMitchell:
The CPU tests here ran for 5 hours hmmmmm.

RAMitchell:
The gtests crash is not related to this PR but reproduces on main with recent versions of legate: [#2239](https://github.com/nv-legate/legate.internal/issues/2239)

python/tests/test_table.py
```diff
@@ -10,6 +10,7 @@
 from legate_dataframe.testing import assert_frame_equal, guess_available_mem
 
 
+@pytest.skip(reason="This causes CI hangs. Investigate rewriting this test.")
```
seberg, line 13:
```suggestion
@pytest.mark.skip(reason="This causes CI hangs. Investigate rewriting this test.")
```
(Although there is another one in `test_column.py` which we probably also need.)

Issue #52, opened 2025-05-05
Prototype CPU support


RAMitchell:
Legate-dataframe cannot currently run on CPU as it lacks CPU implementations for tasks. libarrow can be used to implement CPU functionality (e.g. csv reading) in the same way libcudf is currently used. libarrow is already a dependancy of legate-dataframe.

A first prototype might contain:
- The ability to convert from a legate-dataframe column to arrow::Array and back.
- 1 task (e.g. read_csv) implmented using libarrow
- A test for this in CI, with a CPU only runner

PR #53, opened 2025-05-05
Implement CPU task for read_csv

RAMitchell:
## Description
Implements #52


RAMitchell:
I've added a refactoring commit here that touches 29 files, which I should explain:

legate-dataframe gpu_tasks all begin by creating a GPUTaskContext, with is used to unpack arguments and also carries a GPU memory resource for the task.

This cannot be used in a cpu_task, so I have here seperated the GPU part and relabeled it TaskContext. Instead a gpu_task must seperately create a TaskMemoryResource and pass it into any function or method that needs it (typically some libcudf function). This is a little verbose due to the number of functions/methods that need this, but it makes the PhysicalColumn object itself not depend on a GPU related code - only specific methods or functions.

RAMitchell:
This PR now passes numeric CSV read write tests:
`LEGATE_CONFIG="--gpus 0 --cpus 1" ./cpp/build/gtests/cpp_tests --gtest_filter=*NumericCSV*`

The CSV reader currently allocates 1 file to each worker and cannot yet distribute the work evenly.

Still to do:
- strings
- A CPU only CI test

python/legate_dataframe/cmake/Modules/LinkPyarrowHeaders.cmake
```diff
@@ -15,26 +15,35 @@ include_guard(GLOBAL)
...
-execute_process(
-  COMMAND "${Python_EXECUTABLE}" -c "import pyarrow; print(pyarrow.get_include())"
-  OUTPUT_VARIABLE PYARROW_INCLUDE_DIR
-  ERROR_VARIABLE PYARROW_ERROR
-  RESULT_VARIABLE PYARROW_RESULT
-  OUTPUT_STRIP_TRAILING_WHITESPACE
-)
+# Add pyarrow and numpy includes to each target
+function(link_to_pyarrow_headers targets)
+  execute_process(
```
RAMitchell, line 20: I modified this function because it failed when called a second time. I needed to regenerate the PYARROW_INCLUDE_DIR variable on the second call.

cpp/include/legate_dataframe/core/allocator.hpp
```diff
@@ -104,34 +104,5 @@ class TaskMemoryResource : public DeviceMemoryResource {
...
- * This resource returns allocations backed by either legate buffers or regular CUDA
- * buffers depending on whether the allocation is called from inside a legate task
- * or by the control code. Therefore, this resource can be used both inside and
- * outside tasks and can be used with RMM's `set_per_device_resource()` and
- * `set_current_device_resource()`.
- *
- * Recommendation: use `.set_as_default_mmr_resource()` to create a single instance of
- * this class per GPU and set them as the default memory resources.
- */
-class GlobalMemoryResource : public DeviceMemoryResource {
```
seberg, line 119: @madsbk I forgot to ask. My feeling is that this may be important for libcudf calls, because not everything uses the provided mr(), but I am not sure where it might be relevant or how much.

Do you have a quick feel/know that this is important, or is it likely unnecessary complexity?
(I think restoring this shouldn't be too bad, since we can skip calling it when there are no GPUs in the system.  But, if we are unsure, also OK to restore it in a followup.)

cpp/src/parquet.cpp
```diff
@@ -129,7 +131,7 @@ class ParquetRead : public Task<ParquetRead, OpCode::ParquetRead> {
       for (const auto& table : tables) {
         table_views.push_back(table->view());
       }
-      tbl_arg.move_into(cudf::concatenate(table_views, ctx.stream(), ctx.mr()));
+      tbl_arg.move_into(cudf::concatenate(table_views, context.get_task_stream()));
```
seberg, line 134: OK, too many to suggest.  All of these can remain unchanged now, since we still have the `ctx.stream()` short-hand and `ctx.mr()` is available again and must still be passed in.

cpp/src/utils.cpp
```diff
@@ -246,6 +246,126 @@ legate::Type to_legate_type(cudf::type_id dtype)
...
+      return legate::float64();
+    }
+    case arrow::Type::STRING: {
+      return legate::string_type();
+    }
+    default: throw std::invalid_argument("unsupported Arrow datatype");
+  }
+}
+
+std::shared_ptr<arrow::DataType> to_arrow_type(cudf::type_id code)
```
seberg, line 292: This looks fine.  Do we actually ever use it without having a `cudf::data_type`, though?

I think I would slightly prefer to just pass `const cudf::data_type &cudf_type`, since I suspect there will be (small/odd) places where the type code may not be enough and I doubt there are many places where we only have the ID.

python/legate_dataframe/cmake/Modules/LinkPyarrowHeaders.cmake
seberg: Probably vendored from somewhere, but the changes look fine and easy enough, thanks.

python/legate_dataframe/lib/core/column.pyx
```diff
@@ -106,6 +109,29 @@ cdef class LogicalColumn:
                 "from_cudf() only supports cudf columns and device scalars."
             )
 
+    @staticmethod
+    def from_arrow(array) -> LogicalColumn:
+        """Create a logical column from a local arrow array.
+
+        This call blocks the client's control flow and scatter
```
seberg, line 116: I guess technically, we'll have a local copy on all nodes due to control replication :).

Not sure, how to adapt, just delete the second part or is it more right than it seems to me?

python/legate_dataframe/lib/core/column.pyx
```diff
@@ -106,6 +109,29 @@ cdef class LogicalColumn:
...
+        ----------
+        array
+            pyarrow array
+
+        Returns
+        -------
+            New logical column
+        """
+        print(type(array))
+        print(array)
```
seberg, line 129:
```suggestion
```

python/legate_dataframe/lib/core/column.pyx
```diff
@@ -106,6 +109,29 @@ cdef class LogicalColumn:
                 "from_cudf() only supports cudf columns and device scalars."
             )
 
+    @staticmethod
+    def from_arrow(array) -> LogicalColumn:
```
seberg, line 113:
```suggestion
    def from_arrow(pa.Array array not None) -> LogicalColumn:
```
Maybe?  Then Cython will give a nice error for the type.  (But it seems arrow handles bad types fine, so also OK.)

DEVELOPER_GUIDE.md
```diff
@@ -95,15 +95,13 @@ LogicalColumn unary_operation(const LogicalColumn& col, cudf::unary_operator op)
...
+Always use the CUDA stream from `context->get_task_stream()` and a `TaskMemoryResource` local to the task. This is because Legate may run multiple tasks on the same GPU and calls to CUDA functions such as `cudaMalloc()` might block **all** CUDA kernels on the same device.
```
seberg, line 104: Should we just restore the old text?  (Or maybe just the first part).  `stream()` and `mr()` are back, so that didn't change.

(Modulo a note that this is of course only valid for GPU variants.)

build.sh
```diff
@@ -18,13 +18,14 @@ ARGS=$*
...
 
-VALIDARGS="clean liblegate_dataframe legate_dataframe test -v -g -n -s --ptds -h"
+VALIDARGS="clean liblegate_dataframe legate_dataframe test -v -g -n -s --ptds -h -e"
 HELP="$0 [clean] [liblegate_dataframe] [legate_dataframe] [legate] [-v] [-g] [-n] [-s] [--ptds] [--cmake-args=\"<args>\"] [-h]
    clean                       - remove all existing build artifacts and configuration (start over)
    liblegate_dataframe         - build and install the liblegate_dataframe C++ code (tests are build only)
    legate_dataframe            - build and install the legate_dataframe Python package
    test                        - test all of legate-dataframe (requires valgrind and compute-sanitizer)
    -v                          - verbose build mode
+   -e                          - build in editable mode (python only)
```
seberg, line 28: :tada:

cpp/src/utils.cpp
RAMitchell: Will overhaul the type conversions in a later PR.

python/legate_dataframe/lib/core/column.pyx
RAMitchell: Oops

cpp/src/core/column.cu
```diff
@@ -220,11 +336,51 @@ std::unique_ptr<cudf::column> LogicalColumn::get_cudf(rmm::cuda_stream_view stre
...
+      "Cannot call `.arrow_array()` on a unbound LogicalColumn, please bind it using "
+      "`.move_into()`");
+  }
+  if (array_->nested()) {
+    if (array_->type().code() == legate::Type::Code::STRING) {
+      const legate::StringPhysicalArray a = array_->get_physical_array().as_string_array();
+      const legate::PhysicalArray chars   = a.chars();
+      const auto num_chars                = chars.data().shape<1>().volume();
+
+      std::shared_ptr<arrow::Buffer> data = *arrow::AllocateBuffer(num_chars * sizeof(int8_t));
```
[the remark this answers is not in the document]
RAMitchell: No I mean your helper works nicely! I have switched to using it.

ci/run_ctests_gpu.sh
```diff
@@ -13,6 +13,6 @@ fi
 
 # Unless `LEGATE_CONFIG` is set, default to all available GPUs and set fbmem/sysmem.
 # LEGATE_TEST=1 to test broadcasting code paths (locally).
-LEGATE_CONFIG=${LEGATE_CONFIG:- --gpus="$(nvidia-smi -L | wc -l)" --fbmem=2000 --sysmem=6000 --omps=0} \
+LEGATE_CONFIG=${LEGATE_CONFIG:- --gpus="$(nvidia-smi -L | wc -l)" --sysmem=6000 --omps=0} \
```
seberg, line 16: Sorry, hand't noticed this before. Was there much of a reason to remove this?  (Not that it should really matter).

seberg (approved):
Thanks, let's give this a shot (one tiny comment, but doesn't matter).

We should possibly open some follow up issue.  I am not sure I remember what it was beyond likely restoring the global memory resource.

PR #56, opened 2025-06-17
Implement binary ops with arrow

RAMitchell:


cpp/src/core/column.cu
```diff
@@ -236,12 +236,17 @@ legate::LogicalArray from_arrow(std::shared_ptr<arrow::Array> arrow_array)
   }
   auto array = runtime->create_array({std::uint64_t(arrow_array->length())},
                                      to_legate_type(arrow_array->type_id()),
-                                     1,
-                                     arrow_array->null_bitmap_data() != nullptr);
+                                     arrow_array->null_count() > 0,
```
seberg, line 239: Let's try this! (I am slightly unsure if null count 0 and nullable are subtly different, but columns should be immutable, so...)

cpp/src/reduction.cpp
```diff
@@ -383,8 +383,7 @@ class AggregationHelper final {
         auto cudf_int64 = cudf::data_type{cudf::type_id::INT64};
         auto counts     = get_result(*cudf::make_count_aggregation(), cudf_int64, std::nullopt);
 
-        return legate::dataframe::binary_operation(
-          sum, counts, cudf::binary_operator::DIV, output_type);
+        return legate::dataframe::binary_operation(sum, counts, "div", output_type);
```
seberg, line 386:
```suggestion
        return legate::dataframe::binary_operation(sum, counts, "divide", output_type);
```

cpp/tests/test_binaryop.cpp
```diff
@@ -14,86 +14,164 @@
...
-TYPED_TEST_SUITE(BinaryOpsTest, cudf::test::NumericTypes);
+// Edge cases for unsupported operations or inconsistent implementations between cudf/arrow
+bool skip(std::string const& op, legate::Type const& type)
+{
+  bool use_cudf =
+    legate::Runtime::get_runtime()->get_machine().count(legate::mapping::TaskTarget::GPU) > 0;
+  if (use_cudf) {
+    // Cudf integer power appears broken
+    // https://github.com/rapidsai/cudf/issues/10178#issuecomment-3004143727
+    if (op == "power" && type.to_string().find("int") != std::string::npos) { return true; }
```
seberg, line 41:
```suggestion
    if (op == "power" && cudf::is_integral(type)) { return true; }
```
with `#include <cudf/utilities/traits.hpp>`

cpp/tests/test_binaryop.cpp
```diff
@@ -14,86 +14,164 @@
...
+                                              legate::Type::Code::COMPLEX64,
+                                              legate::Type::Code::COMPLEX128};
+  if (float_ops.count(op) && !float_types.count(type.code())) { return true; }
+
+  // We don't support bitwise operations on floating point types
+  std::set<std::string> bitwise_ops = {
+    "bit_wise_and", "bit_wise_or", "bit_wise_xor", "shift_left", "shift_right"};
+  if (bitwise_ops.count(op) && float_types.count(type.code())) { return true; }
+
+  return false;
```
seberg, line 72: ufff, but looks good.

cpp/tests/test_binaryop.cpp
```diff
@@ -14,86 +14,164 @@
...
+                                                 arrow::compute::CastOptions::Unsafe()))
+                 .make_array();
+    auto result = binary_operation(lhs, rhs, op, lhs.cudf_type()).get_arrow();
+
+    // For integers check exact equality, for floats check approximate equality
+    std::set<legate::Type::Code> float_types = {legate::Type::Code::FLOAT32,
+                                                legate::Type::Code::FLOAT64,
+                                                legate::Type::Code::FLOAT16,
+                                                legate::Type::Code::COMPLEX64,
+                                                legate::Type::Code::COMPLEX128};
```
seberg, line 106: Could use `is_integral` here also, but also happy to keep as is.

cpp/tests/test_binaryop.cpp
```diff
@@ -14,86 +14,164 @@
...
 
-  CUDF_TEST_EXPECT_COLUMNS_EQUAL(res.get_cudf()->view(), expect->view());
+TYPED_TEST(BinaryOpsTest, AddColCol)
+{
+  LogicalColumn lhs(narrow<TypeParam>({1, 2, 3, 4}));
+  LogicalColumn rhs(narrow<TypeParam>({5, 6, 7, 8}));
+  CompareArrow(lhs, rhs, ops);
 }
 
 TYPED_TEST(BinaryOpsTest, AddColColWithNull)
```
seberg, line 151: Maybe rename, since it's a lot of ops now.

python/legate_dataframe/lib/core/data_type.pyx
```diff
@@ -26,6 +28,11 @@ cdef cpp_cudf_type as_data_type(data_type_like):
         Coerced C++ cudf type.
     """
     cdef DataType d_type
+
+    # arrow
+    if isinstance(data_type_like, pa.DataType):
+        data_type_like = from_arrow(data_type_like)
```
seberg, line 34: nice!  Would be good to add to the `.pyi`.

python/legate_dataframe/lib/core/scalar.pyx
```diff
@@ -40,6 +41,9 @@ cdef LogicalColumn cpp_scalar_col_from_python(scalar: ScalarLike):
     if isinstance(scalar, legate.core.Scalar):
         scalar = scalar.value()
 
+    if isinstance(scalar, pa.Scalar):
+        return LogicalColumn.from_arrow(scalar)
```
seberg, line 45: Maybe remove the typing here, but add `pa.Scalar` to the `.pyi`.  (My bad for duplicating it here.)

seberg (commented):
Very nice generally, too bad that a few of the ops are not tested.

We could add some failure tests (e.g. bad operator string) just to see that it fails early, but if that is annoying, then happy not to.

Really just smaller comments.

python/legate_dataframe/lib/core/scalar.pyx
seberg: The `scalar.pyi` defines `ScalarLike`.  It makes no sense to keep `ScalarLike` around here, but it makes sense to include `pa.Scalar` there.

RAMitchell:
This now supports arrow boolean arrays and tests comparison operators.

seberg (approved):
New fixes/changes look good. happy to ship if tests pass, thanks!

PR #57, opened 2025-06-18
Add `LDF_PREFER_EAGER_ALLOCATIONS` for legate investigations

seberg:
Legate folks asked a way to use eager allocations more.  This might actually be useful for in some places (or at least the machinery around it).

I didn't set up the tests, I ran them locally with 2 GPUs and they run fine except, that using this for joins is actually incorrect so one tests (correctly) fails.

To use just run with `LDF_PREFER_EAGER_ALLOCATIONS=1`, and many ops will use eager allocations:
* to/from timestamp
* casts
* unary and binary ops
* replace_nulls
* Join...
  * If LEFT join
  * and explicit right side broadcasting
* ~parquet reading~ (removed in this PR for now)

(As said, even then it is incorrect: a left join can return a different sized result if there are duplicates in the RHS.  Doing this will be flagged as an error, and can lead to segfaults.)

---

I would be OK with putting this in if it helps for whatever tries someone wants to do.  I mean it if I say that the general ability may actually be useful even if the environment variable is clearly wrong with applying it to join.
It does add a bit of cruft, though.

### DIFF
--- a/cpp/include/legate_dataframe/core/column.hpp
+++ b/cpp/include/legate_dataframe/core/column.hpp
@@ -210,22 +210,14 @@ class LogicalColumn {
    * must use the long signature.
    *
    * @param other The prototype column
-   * @param bound Whether the new column should be bound with the size of the input.
    * @return The new unbounded column with the type and nullable equal `other`
    */
-  static LogicalColumn empty_like(const LogicalColumn& other, bool bound = false)
+  static LogicalColumn empty_like(const LogicalColumn& other)
   {
-    if (!bound) {
-      return LogicalColumn(legate::Runtime::get_runtime()->create_array(
-                             other.array_->type(), other.array_->dim(), other.array_->nullable()),
-                           other.cudf_type(),
-                           false);
-    } else {
-      return LogicalColumn(legate::Runtime::get_runtime()->create_array(
-                             other.array_->shape(), other.array_->type(), other.array_->nullable()),
-                           other.cudf_type(),
-                           true);
-    }
+    return LogicalColumn(legate::Runtime::get_runtime()->create_array(
+                           other.array_->type(), other.array_->dim(), other.array_->nullable()),
+                         other.cudf_type(),
+                         false);
   }
 
   /**
@@ -479,15 +471,18 @@ class PhysicalColumn {
   /**
    * @brief Indicates whether the column is unbound or not
    *
+   * For string columns the underlying characters will be (or can be) unbound
+   * even if the ranges are bound.
+   *
    * @return true The column is unbound
    * @return false The column is bound
    */
   [[nodiscard]] bool unbound() const
   {
-    // If one of the underlying stores are unbound, the column as a whole is unbound.
-    // TODO: cache this value
-    const std::vector<legate::PhysicalStore> ss = get_stores(array_);
-    return std::any_of(ss.cbegin(), ss.cend(), [](const auto& s) { return s.is_unbound_store(); });
+    if (!array_.nested()) { return array_.data().is_unbound_store(); }
+    // New string columns won't really have bound characters so only check ranges.
+    // (only nested dtypes we have right now)
+    return array_.as_string_array().ranges().data().is_unbound_store();
   }
 
   /**
--- a/cpp/src/core/column.cu
+++ b/cpp/src/core/column.cu
@@ -619,7 +619,7 @@ struct move_into_fn {
                   std::unique_ptr<cudf::column> column,
                   cudaStream_t stream)
   {
-    // The string version currently doesn't support already bound outputs.  Presumably, this
+    // The string version currently doesn't support already bound chars outputs.  Presumably, this
     // can't happen right now anyway, because the result size is not fixed so it should fail early?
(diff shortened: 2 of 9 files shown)


cpp/src/core/column.cu
```diff
@@ -142,7 +142,9 @@ T* maybe_bind_buffer(legate::PhysicalStore store, std::size_t size)
   if (store.is_unbound_store()) {
     out = store.create_output_buffer<T, 1>(legate::Point<1>(size), true).ptr(0);
   } else {
-    out = store.write_accessor<T, 1>().ptr(0);
+    auto acc = store.write_accessor<T, 1>();
+    assert(store.shape<1>().hi[0] == -1 || acc.accessor.is_dense_row_major(store.shape<1>()));
+    out = acc.ptr(store.shape<1>().lo[0]);
```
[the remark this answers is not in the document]
RAMitchell: Yes looks like a bug.

cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -253,12 +271,22 @@ class LogicalColumn {
    * @param scalar Whether the result is a scalar column.
    * @return The new unbounded column
    */
-  static LogicalColumn empty_like(cudf::data_type dtype, bool nullable, bool scalar = false)
+  static LogicalColumn empty_like(cudf::data_type dtype,
```
RAMitchell, line 274: Do we need all of these empty_like constructors?

cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -603,27 +631,33 @@ class PhysicalColumn {
...
    * @param column The arrow array to move
+   * @param allow_copy If true, copies the data if the column is bound.
    */
-  void move_into(std::shared_ptr<arrow::Array> column);
+  void move_into(std::shared_ptr<arrow::Array> column, bool allow_copy = false);
 
   /**
    * @brief Makes the unbound column empty. Valid only when the column is unbound.
+   *
+   * @param allow_copy If true, no error is given if the column is bound, this becomes
```
RAMitchell, line 657: This comment is more descriptive than the above, maybe duplicate it there.

cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -603,27 +631,33 @@ class PhysicalColumn {
    * @brief Move local cudf column into this unbound physical column
    *
    * @param column The cudf column to move
+   * @param allow_copy If true, copies the data if the column is bound.
    */
-  void move_into(std::unique_ptr<cudf::column> column);
+  void move_into(std::unique_ptr<cudf::column> column, bool allow_copy = false);
```
RAMitchell, line 636: So move into will lazily try to move the input if it can, and otherwise copy into any existing buffer?

Is there any argument for having a move_into and copy_into function to make this explicit?

cpp/src/core/library.cpp
```diff
@@ -153,4 +157,13 @@ legate::Library& get_library()
   return library;
 }
 
+bool get_prefer_eager_allocations()
```
RAMitchell, line 160: This is not intended to be a long term solution right?

RAMitchell (approved):
If I understand correctly, using eager allocations comes at the cost of extra memory, as the cudf/arrow result cannot be reused by moving into the legate frame.

For something like a binary op we could hypothetically implment it ourselves without arrow/cudf and have the best of both worlds. Just a thought.

cpp/include/legate_dataframe/core/column.hpp
seberg: Yeah, you are right, that is nicer API wise.  I tried to avoid branching in the tasks for now.  But I guess a:
```
if get_():
    tbl.copy_into();
else:
    tbl.move_into();
```
at the end of a few tasks isn't bad since it is likely not quite permanent anyway.

cpp/src/core/library.cpp
RAMitchell: IMO, from a user perspective, users shouldn't to have to touch any parameters for streaming to work - it should just work. This is a demo so its fine for now.

seberg:
@RAMitchell thanks for the review. I added `copy_into()` and `move_into()`.  Unfortunately that (and a bit removing the `empty_like()` lead to some follow-ups and slight re-org for the `copy_into()`/`move_into()` branching:
* Tweak the definition of `unbound()`, but I think it is just more useful even if it's a bit unclear what it means for string arrays.
* Add `copy_to`, I made the `move_into()` now call a new `from_cudf()` to mirrow the `from_arrow()` helper and moved most error checks there just to not duplicate them.

So if/when you have a moment, may make sense to browse through once more.

RAMitchell (approved):

seberg:
> I wonder if its easier when we just use eager allocations everywhere possible, preferring the ability to stream over temporary memory usage in tasks.

Yeah, although there are is important functionality where eager allocation is just not possible, so I am not sure it's worth transitioning to it wherever it is.  (But, although may well be if it helps legate with other things.)

PR #59, opened 2025-06-25
ENH: Read parquet files by row group

seberg:
This is a bit of a story to get there with two steps forward and one backwards.

I would like to use image constraint, but legate does not seem to properly support them here (I have no idea why).
Thus, this goes "back" to using a late-bound store (half the idea of a "read array" was to NOT do this but there is no helping it).

I did keep a decent chunk of the image constraint approach around even if unused.

Which also means having to refactor/duplicate some of the copying code because we have to work with buffers and can't use a `write_accessor` on late bound stores.

This should make reading parquet files faster and more robust but of course limits the partitioning to row-group multiples, which is probably good for temproraries, but if row groups are huge and don't fit into GPUs evently, it isn't necessarily ideal.

## Checklist
- [x] Run `./build.sh test` for local testing, see [CONTRIBUTING.md](https://github.com/rapidsai/legate-dataframe/blob/main/CONTRIBUTING.md#testing).

### DIFF
--- a/cpp/include/legate_dataframe/core/transposed_copy.cuh
+++ b/cpp/include/legate_dataframe/core/transposed_copy.cuh
@@ -34,7 +34,6 @@ void copy_into_tranposed(TaskContext& ctx,
                          void* data_ptr,
                          std::optional<bool*> null_ptr,
                          cudf::table_view tbl,
-                         size_t offset,
                          legate::Scalar& null_value);
 
 }  // namespace legate::dataframe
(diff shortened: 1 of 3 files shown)


seberg:
Thanks, let's give this a shot then.  Once https://github.com/nv-legate/legate.internal/pull/2413 is in (and available) we can hopefully revert the work-around for the non-working image constraints for the array reader.

(At that point, maybe we should consider also using it for the normal one, even if it probably means doing an internal copy.)

PR #61, opened 2025-06-27
MAINT: Bump to legate 25.07 and RAPIDS 25.04

seberg:
This bumps all versions with some bumps along the road.

I made two changes that slightly affect API:
1. Accepting now all "column like" data and all "scalar like" that cudf/pylibcudf have to offer.  Just for convenience really and it's not a big thing.
2. `DeviceScalar` is gone, so returning a `cudf.Scalar`.

It may make sense to add `to_pylibcudf()` (and `to_pylibcudf_scalar()` and maybe `from_pylibcudf()` even if I made the `from_cudf()` versions accept those).
But doesn't seem necessary here, and I am not sure it is very relevant at all for now.

### DIFF
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -46,7 +46,6 @@ jobs:
         CUDA_VER:
           - "12.5.1"
         PY_VER:
-          - "3.10"
           - "3.11"
           - "3.12"
     runs-on: linux-${{ matrix.ARCH }}-cpu4
--- a/.github/workflows/docs-build.yaml
+++ b/.github/workflows/docs-build.yaml
@@ -30,7 +30,7 @@ jobs:
         include:
           - ARCH: amd64
             CUDA_VER: "12.5.1"
-            PY_VER: "3.11"
+            PY_VER: "3.12"
     runs-on: linux-${{ matrix.ARCH }}-gpu-v100-latest-1
     container:
       image: "rapidsai/ci-conda:cuda${{ matrix.CUDA_VER }}-ubuntu22.04-py${{ matrix.PY_VER }}"
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -50,7 +50,7 @@ jobs:
         include:
           - ARCH: "amd64"
             CUDA_VER: "12.5.1"
-            PY_VER: "3.10"
+            PY_VER: "3.11"
           - ARCH: "amd64"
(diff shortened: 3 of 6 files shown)


ci/run_ctests_cpu.sh
```diff
@@ -14,4 +14,4 @@ fi
 # LEGATE_TEST=1 to test broadcasting code paths (locally).
 LEGATE_CONFIG=${LEGATE_CONFIG:- --cpus 8 --gpus 0 --sysmem=6000 --omps=0} \
 LEGATE_TEST=${LEGATE_TEST:-1} \
-legate ./cpp_tests --output-on-failure --no-tests=error "$@" --gtest_filter="*CSV*"
+legate ./cpp_tests --output-on-failure --no-tests=error "$@" --gtest_filter="*CSV*:*Binary*"
```
seberg, line 17: I don't know what coolaid github is drinking here to include this diff from the previous PR (happened yesterday as well).

cpp/CMakeLists.txt
```diff
@@ -121,6 +126,11 @@ set_target_properties(
              INTERFACE_POSITION_INDEPENDENT_CODE ON
 )
 
+# Need to add this define, as is done in legate CMakeLists.txt as well for __half support.  If
+# __half support fails elsewhere it may be needed there. This may be a CCCL 2.7.0 issue and become
+# unnecessary in the future.
+target_compile_definitions("LegateDataframe" PUBLIC _LIBCUDACXX_HAS_NVFP16=1)
```
seberg, line 132: @RAMitchell this is something to remember when doing this for legate-boost. (Also the CCCL thing likely, while it makes sense here, it is a bit surprising that it picked up a random CCCL version rather than one from legate, since I think legate was the first dependency here.)
The CCCL was a bit confusing, because in some versions even when it claimed to use 2.7, it wasn't.

ci/run_ctests_cpu.sh
seberg: Nvm, something was wrong with the rebase, the last merge commit was there, but it was incorrect.

cpp/CMakeLists.txt
RAMitchell: Noted thanks!

RAMitchell (approved):

seberg:
Had to remove Python 3.10 support without adding 3.13, but I don't think it's a big enough change.  So looks like this is OK now, putting it in.

PR #62, opened 2025-06-27
Parquet read/write cpu tasks

RAMitchell:


RAMitchell:
Unfortunately `schema->field(i)->nullable()` which is being used in the parquet reader to decide if the legate array should be nullable is pretty much useless: https://stackoverflow.com/q/78848723. It seems to just always be true.

cpp/src/parquet.cpp
```diff
@@ -48,6 +49,27 @@ class ParquetWrite : public Task<ParquetWrite, OpCode::ParquetWrite> {
...
+
+    const std::string dirpath  = argument::get_next_scalar<std::string>(ctx);
+    const auto column_names    = argument::get_next_scalar_vector<std::string>(ctx);
+    const auto table           = argument::get_next_input<PhysicalTable>(ctx);
+    const std::string filepath = dirpath + "/part." + std::to_string(ctx.rank) + ".parquet";
+    auto outfile               = ARROW_RESULT(arrow::io::FileOutputStream::Open(filepath));
+    auto props                 = parquet::WriterProperties::Builder().build();
+    auto arrow_props           = parquet::ArrowWriterProperties::Builder().build();
+
+    // TODO: memory pool should come from legate
```
seberg, line 64: Yeah, happy to leave the memory pool out of here.

cpp/src/parquet.cpp
```diff
@@ -124,11 +146,77 @@ struct create_result_store_fn {
...
+  // Find the column indices
+  std::vector<int> column_indices;
+  column_indices.reserve(columns.size());
+  for (const auto& col : columns) {
+    auto idx = schema->GetFieldIndex(col);
+    if (idx < 0) { throw std::runtime_error("Column " + col + " not found in Parquet schema."); }
+    column_indices.push_back(idx);
+  }
+  return column_indices;
+}
```
seberg, line 170: Should we maybe just add that information in the prep step already and pass it in even if the gpu task ends up not using it? But this seems OK also and includes automatically raising an error on mismatch.

cpp/src/parquet.cpp
```diff
@@ -124,11 +146,77 @@ struct create_result_store_fn {
...
+      auto status = parquet::arrow::OpenFile(input, arrow::default_memory_pool(), &arrow_reader);
+      std::unique_ptr<arrow::RecordBatchReader> batch_reader;
+      status = arrow_reader->GetRecordBatchReader(row_groups[i], column_indices, &batch_reader);
+      tables.push_back(ARROW_RESULT(batch_reader->ToTable()));
+    }
+    // Concatenate the tables
+    if (tables.size() == 0) {
+      tbl_arg.bind_empty_data();
+    } else {
+      tbl_arg.move_into(std::move(ARROW_RESULT(arrow::ConcatenateTables(tables))));
```
seberg, line 216: I do wonder if it's worth adding a `move_into()` (if we go ahead with something like the other PR) that does the concatenation part and copies exactly once.

That way, we could use an eager allocation, and just always copy.  And we could even read one row group at a time and then copy (or similar).

Anyway, nothing to do here!

python/legate_dataframe/testing.py
```diff
@@ -50,7 +50,18 @@ def as_cudf_dataframe(obj: Any, default_column_name: str = "data") -> cudf.DataF
 
 
 def assert_arrow_table_equal(left: pa.Table, right: pa.Table) -> None:
-    assert left.equals(right), f"Arrow tables are not equal:\n{left}\n{right}"
+    # arrow has an annoying nullable attribute in its schema that is not well respected by its various functions
+    # i.e. it is possible to have a non-nullable column with null values without problems
+    # Set the nullable attribute to match the left table
```
seberg, line 55: Hmmm, this seems unfortunate, how bad is it?  I don't really want to try to see if we can just live with and make sure to keep it happy.

(I can see a "nullable" column that has definitely no null mask, the opposite doesn't really make sense to me as it would be an incorrect hint at best...)

python/tests/test_parquet.py
```diff
@@ -13,101 +13,132 @@
...
+    assert_arrow_table_equal,
+    assert_frame_equal,
+    std_dataframe_set_cpu,
+)
+
+
+def write_partitioned_parquet(table, path, npartitions=1):
+    if table.num_rows == 0:
+        pq.write_table(table, f"{path}/part-0.parquet")
+    partition_size = int(math.floor(table.num_rows / npartitions))
```
seberg, line 38: I guess you initially might had a `ceil` here, just use:
```suggestion
    partition_size = table.num_rows // npartitions
```

(Feels like there should be a trick to do this shorter, but it's not worth searching :))

python/tests/test_parquet.py
```diff
@@ -123,6 +154,9 @@ def test_read_array(tmp_path, npartitions=2, glob_string="/*"):
     assert arr.dtype == arr_from_tbl.dtype
     assert cn.array_equal(arr_from_tbl, arr)
 
+    assert arr.dtype == arr_from_tbl.dtype
+    assert np.array_equal(arr_from_tbl, arr)
+
```
seberg, line 159:
```suggestion
```
Or was there a reason, seems just a repeat of the above.

(Unless we want to write this with only numpy and avoid cupynumeric, but not sure if legate supports that for its logical array.)

cpp/src/core/transposed_copy.cu
```diff
@@ -169,4 +170,95 @@ void copy_into_tranposed(TaskContext& ctx,
...
+                               .column_idx  = i,
+                               .num_columns = static_cast<std::size_t>(table->num_columns()),
+                               .row_offset  = row_offset};
+
+      // Cast if necessary
+      auto target_arrow_type = to_arrow_type(to_cudf_type_id(type.code()));
+      if (array->type_id() != target_arrow_type->id()) {
+        auto casted_array = ARROW_RESULT(arrow::compute::Cast(*array, target_arrow_type));
+        array             = std::move(casted_array);
+      }
```
seberg, line 252: Feel a bit like if we do that here for the CPU, we should also move it for the GPU?

(The annoying thig is that, we really want the `array` version here, the other version is just a work-around for not using eager allocations.)

seberg (commented):
Looks good, thanks!

I wish we could figure out a way to reduce the duplication a bit more...

The test failure is real, although in the GPU version:
```
LEGATE ERROR: #0 Task "legate::dataframe::task::ParquetReadArray" threw an unexpected exception (legate::detail::TracedException<std::invalid_argument>): "LEGATE ERROR: ================================================================================
LEGATE ERROR: System: Linux, 6.8.0-60-generic, 164ab292dfe4, #63-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 19:04:15 UTC 2025, x86_64
LEGATE ERROR: Legate version: 25.5.0 (552803cfc816f383d377ab673b23bdd4cb371249)
LEGATE ERROR: Legion version: 25.5.0 (04b7d5068c5e75f29684703e8a1b8568b3e59b9a)
LEGATE ERROR: Configure options: --LEGATE_ARCH=arch-conda --with-python --with-cc=/tmp/conda-croot/legate/_build_env/bin/x86_64-conda-linux-gnu-cc --with-cxx=/tmp/conda-croot/legate/_build_env/bin/x86_64-conda-linux-gnu-c++ --build-march=x86-64 --cmake-generator=Ninja --legion-max-dim=6 --with-openmp --with-cuda --with-cal --build-type=release --with-ucx
LEGATE ERROR: Exception stack contains 1 exception(s) (bottom-most exception first):
LEGATE ERROR:
LEGATE ERROR: #0 std::invalid_argument: Size of the scalar is 0, but the requested type has size 1
```
may need to compile with debugging to see where that comes from.  It looks like the task argument unpacking to me, but that doesn't even differ between GPU and CPU, so I find it rather surprising?

cpp/src/core/transposed_copy.cu
RAMitchell: It may not be necessary.

python/legate_dataframe/testing.py
seberg: TL;DR: OK, doesn't seem to bad, and I don't want to distinguish "nullable" and "has no nulls" really.

Hmmm, OK, that does look tricky. I guess a nullable field may or may not contain nulls, but a non-nullable one can get them but is inconsistent and causes bugs.

In principle, it could make sense for us to distinguish nullable from "contains no nulls" which feels like it may resolve _some_ of this.
But, our columns are immutable and I don't think libcudf does this distinction either (OK, only in the sense of a nullable column with zero nulls, but the null mask is allocated).

In practice, I agree that when reading parquet to an array at least, it would be more pragmatic to check if there are actual nulls (rather than trusting the `nullable`).  That information is likely (often) available in the metadata...
... But in practice, while reading metadata may make sense as a query optimization step eventually, I don't really feel it makes a lot of sense here.

cpp/src/core/transposed_copy.cu
RAMitchell: It is necessary nvm.

RAMitchell:
The test is failing because the cudf transpose code attempts to access a scalar and this optional argument is not given by the user.

PR #63, opened 2025-06-27
Fixup parquet chunking

seberg:
The old value didn't make sense, this should be in bytes. (I doubt this makes a difference, since the old value should just have been too small.)

https://docs.rapids.ai/api/libcudf/stable/classcudf_1_1io_1_1chunked__parquet__reader#a49f5549b53257828d50f5fa65114e07a


seberg:
Putting it in, it seems to give reasonable chunks locally with some light testing.

However... now with cudf 25.04 I am hitting other problems that lead to crashes, hangs, and maybe absurd allocations... feels like use of uninitialized memory or so.

PR #65, opened 2025-06-30
ENH: Work towards polars support

seberg:
This is still more rough than I would like, and while I tested locally, I need to add tests (or better, figure out how to integrate existing tests while just allowing many unsupported paths to skip).

Should actually support much of what we need:
* polars reading
* basic CSV reading
* Binary, unary ops, and casts
* joins
* boolean filters.

Not yet, reductions or groupby-aggs and some other things.  And there are most likely many holes.

This is based on `cudf-polars` mostly copied verbatim with minor adjustments, changes are mostly e.g. confined to `do_evaluate`.

---

Right now (I need to see if there is a nicer way).  The main API is:
```
import legate_dataframe.ldf_polars  # needed to monkey-patch for now

query = # polars lazy frame query
query.legate.collect()  # query.collect(engine=...) can probably be done
```
I also added:
```
LogicalTable.lazy()
```
to go from legate-df to a polars `LazyFrame`.
(That part may need a lot of polishing, also since `engine=` may be an alternative path, but if we want to return legate-df tables, this may well be the only good way to do it.)


python/legate_dataframe/ldf_polars/containers/column.py
```diff
@@ -0,0 +1,144 @@
...
+    obj: LogicalColumn
+    # NOTE: cudf_polars stores sortedness, we may need that eventually.
+    is_scalar: bool
+    # Optional name, only ever set by evaluation of NamedExpr nodes
+    # The internal evaluation should not care about the name.
+    name: str | None
+
+    def __init__(
+        self,
+        column: LogicalColumn,
```
RAMitchell, line 36: Would love to see constructors for arrow/cudf for interoperability and testing.

RAMitchell (approved):
Looks good as a starting point. Feel free to merge.

PR #70, opened 2025-07-02
Implment unary ops for CPU

RAMitchell:


python/tests/test_unaryop.py
```diff
@@ -14,36 +14,65 @@
...
+    "invert",
+    "negate",
+]
+
+
+@pytest.mark.parametrize("op", ops)
 def test_unary_operation(op):
-    if op in (UnaryOperator.BIT_INVERT, UnaryOperator.NOT):
-        series = cudf.Series(cupy.random.randint(0, 2, size=1000).astype(bool))
+    if op in ("invert"):
```
seberg, line 52:
```suggestion
    if op in ("invert",):
```

cpp/tests/test_unaryop.cpp
```diff
@@ -31,45 +33,113 @@
...
+    auto expected = (*arrow::compute::CallFunction(op, {col.get_arrow()})).make_array();
+    auto result   = unary_operation(col, op).get_arrow();
+
+    // For integers check exact equality, for floats check approximate equality
+    std::set<legate::Type::Code> float_types = {legate::Type::Code::FLOAT32,
+                                                legate::Type::Code::FLOAT64,
+                                                legate::Type::Code::FLOAT16,
+                                                legate::Type::Code::COMPLEX64,
+                                                legate::Type::Code::COMPLEX128};
+    if (float_types.count(col.type().code())) {
```
seberg, line 95: Feels like it may make sense to move part of this into a utility (soon).

cpp/src/parquet.cpp
```diff
@@ -427,21 +427,14 @@ ParquetReadInfo get_parquet_info(const std::string& glob_string,
...
-
   // Open the first file to get schema information
   auto reader = ARROW_RESULT(arrow::io::ReadableFile::Open(file_paths[0]));
 
-  // Newer versions arrow have versions that return a result which is more convenient (also below)
-  std::unique_ptr<parquet::arrow::FileReader> parquet_reader;
-  auto status = parquet::arrow::OpenFile(reader, pool, &parquet_reader);
-  if (!status.ok()) {
-    throw std::runtime_error("failed to open parquet file: " + status.ToString());
-  }
```
seberg, line 441: Nice, thanks for this cleanup, now that we can do it!

cpp/src/parquet.cpp
```diff
@@ -427,21 +427,14 @@ ParquetReadInfo get_parquet_info(const std::string& glob_string,
...
+  // TODO: Using the default memory pool, hopefully it doesn't matter anyway here.
+  auto parquet_reader =
+    ARROW_RESULT(parquet::arrow::OpenFile(reader, arrow::default_memory_pool()));
   std::shared_ptr<arrow::Schema> schema;
-  status = parquet_reader->GetSchema(&schema);
-  if (!status.ok()) { throw std::runtime_error("failed to get schema: " + status.ToString()); }
+  auto status = parquet_reader->GetSchema(&schema);
```
seberg, line 437: Is there a version now that allows:
```suggestion
  auto schema = ARROW_RESULT(parquet_reader->GetSchema());
```
otherwise, I think the status check still makes sense (if unlikely to fail).

cpp/src/unaryop.cpp
```diff
@@ -45,19 +57,56 @@ class CastTask : public Task<CastTask, OpCode::Cast> {
...
+    {"acosh", cudf::unary_operator::ARCCOSH}, {"atanh", cudf::unary_operator::ARCTANH},
+    {"exp", cudf::unary_operator::EXP},       {"ln", cudf::unary_operator::LOG},
+    {"sqrt", cudf::unary_operator::SQRT},     {"ceil", cudf::unary_operator::CEIL},
+    {"floor", cudf::unary_operator::FLOOR},   {"abs", cudf::unary_operator::ABS},
+    {"round", cudf::unary_operator::RINT},    {"bit_wise_not", cudf::unary_operator::BIT_INVERT},
+    {"invert", cudf::unary_operator::NOT},    {"negate", cudf::unary_operator::NEGATE}};
+
+  if (arrow_to_cudf_ops.find(op) != arrow_to_cudf_ops.end()) { return arrow_to_cudf_ops[op]; }
+  throw std::invalid_argument("Could not find cudf binary operator matching: " + op);
+  return cudf::unary_operator::ABS;
```
seberg, line 81:
```suggestion
```

seberg (approved):
Nice, LGTM.  Some small comments, some tiny of which should be addressed, I think.  But good to go basically.

The test failures are real, though, but should be a simple fixup of just using the string value.

cpp/src/parquet.cpp
RAMitchell: No they didn't update this function in arrow yet.

cpp/src/unaryop.cpp
RAMitchell: The compiler will complain if there is no return type and there is no "null" option given by cudf.

seberg: Ah OK, nvm then. Somehow thought it would be OK behind a throw.

PR #72, opened 2025-07-07
Implement CPU reductions

RAMitchell:


cpp/src/utils.cpp
```diff
@@ -246,9 +248,9 @@ legate::Type to_legate_type(cudf::type_id dtype)
   }
 }
 
-legate::Type to_legate_type(arrow::Type::type code)
+legate::Type to_legate_type(const arrow::DataType& arrow_type)
```
seberg, line 251: Thanks for cleaning this up!

cpp/src/reduction.cpp
```diff
@@ -81,22 +98,67 @@ class ReduceLocalTask : public Task<ReduceLocalTask, OpCode::ReduceLocal> {
...
+        auto sum     = ARROW_RESULT(arrow::compute::Sum(array));
+        result_array = ARROW_RESULT(arrow::MakeArrayFromScalar(*sum.scalar(), 1));
+      }
+    } else {
+      if (initial) {
+        auto initial_col     = argument::get_next_input<PhysicalColumn>(ctx);
+        auto initial_array   = initial_col.arrow_array_view();
+        auto result          = ARROW_RESULT(arrow::compute::CallFunction(op, {array})).scalar();
+        auto result_as_array = ARROW_RESULT(arrow::MakeArrayFromScalar(*result, 1));
+        // Combine two arrays and reduce again
```
seberg, line 128: That's too bad, but if an "initial" isn't explicitly supported, I guess this is the pragmatic thing.  Special casing it for specific operation doesn't seem useful...

cpp/src/reduction.cpp
```diff
@@ -72,6 +73,22 @@ std::unique_ptr<cudf::reduce_aggregation> make_reduce_aggregation(cudf::aggregat
   }
 }
 
+std::unique_ptr<cudf::reduce_aggregation> make_cudf_reduce_aggregation(const std::string& agg_kind)
```
seberg, line 76: `make_reduce_aggregation` seems now dead code, good to just delete, I think.

cpp/tests/test_reduction.cpp
```diff
@@ -30,74 +28,62 @@
...
 
-  auto expected   = cudf::reduce(col, *agg, res_dtype);
-  auto res        = reduce(lg_col, *agg, res_dtype);
-  auto res_scalar = res.get_cudf_scalar();
+  auto expected = ARROW_RESULT(arrow::compute::CallFunction("max", {col.get_arrow()})).scalar();
+  auto expected_array = ARROW_RESULT(arrow::MakeArrayFromScalar(*expected, 1));
 
-  CUDF_TEST_EXPECT_COLUMNS_EQUAL(cudf::make_column_from_scalar(*res_scalar, 1)->view(),
-                                 cudf::make_column_from_scalar(*expected, 1)->view());
+  EXPECT_TRUE(res.get_arrow()->ApproxEquals(expected_array))
```
seberg, line 46:
```suggestion
  EXPECT_TRUE(res.get_arrow()->Equals(expected_array))
```
I suppose it doesn't matter, but seems might as well use exact version.

seberg (approved):
Nice, the initial handling is too bad, but good.  Only real comment, is that we should remove that one function that seems unused now.

I am glad it seemed pretty painless to thread strings in here wasn't quite certain it would be.

---

I just realized that we are in an awkward API position right now, where half of the API is "arrow style" string based, and the other is libcudf style enum based stille (groupby-aggs and joins).

But, at this point I don't think there is anything to do about it for the 25.07 release, unfortunately.

PR #73, opened 2025-07-07
API: Add support for basic slicing

seberg:
Looking through what would be good to have e.g. for TPC-h, slicing is one thing we will probably need because it enables `head()`/`tail()`

(Although, e.g. for sorting the limit should be lowered into the sort itself, since it can be performed on each local partition.)


cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -417,6 +417,15 @@ class LogicalColumn {
    */
   std::string repr(size_t max_num_items = 30) const;
 
+  /**
+   * @brief Slice the column
+   *
+   * @param slice The Legate slice to the column. Note that legate slices are
+   * inclusive and do not support negative values!  TODO: Check!
```
RAMitchell, line 424: TODO here should be done.

cpp/tests/test_slice.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "test_utils.hpp"
+#include <arrow/api.h>
+
+#include <cudf/column/column_view.hpp>
```
RAMitchell, line 20: Remove cudf includes.

cpp/tests/test_slice.cpp
```diff
@@ -0,0 +1,49 @@
...
+  auto expect = col.get_arrow()->Slice(0, 5);  // uses start, length
+  EXPECT_TRUE(expect->Equals(*result.get_arrow()));
+
+  result = col.slice(legate::Slice(1, 3));
+  expect = col.get_arrow()->Slice(1, 2);
+  EXPECT_TRUE(expect->Equals(*result.get_arrow()));
+
+  result = col.slice(legate::Slice(0, -2));
+  expect = col.get_arrow()->Slice(0, 3);  // uses start, length
+  EXPECT_TRUE(expect->Equals(*result.get_arrow()));
```
RAMitchell, line 48: What happens with Slice(0,0)?

RAMitchell: Nvm this is tested in pytests.

RAMitchell (approved):

cpp/include/legate_dataframe/core/column.hpp
seberg: Ooops!, to my surprise, it seems to behave like Python slicing.  (I think that may be a doc issue in legate, since I don't think it's obvious in legate land.)

PR #74, opened 2025-07-08
Implement CPU task for sorting

RAMitchell:


RAMitchell:
A few notes to self:
- Would be good to test with hypothesis or more parameterised pytests
- Test with different numbers of partitions using CPU scoping to expose any issues there
- The comparison I have implemented so far by scratch could instead be an arrow sort over two rows
- Perhaps vectorised lower bound can by implemented by merging the "needle" and "haystack" arrays, sorting and then finding the positions of the "needle" array.

cpp/src/sort.cpp
```diff
@@ -105,6 +331,150 @@ std::unique_ptr<cudf::column> get_split_ind(TaskContext& ctx,
...
+
+template <typename T>
+std::vector<T> column_to_vector(TaskContext& ctx, const cudf::column_view& col)
+{
+  std::vector<T> ret(col.size());
+  if (col.size() > 0) {
+    LEGATE_CHECK_CUDA(cudaMemcpyAsync(
+      ret.data(), col.data<T>(), col.size() * sizeof(T), cudaMemcpyDeviceToHost, ctx.stream()));
+  }
+  return ret;
```
seberg, line 354: Nice, thanks for introducing this helper!

cpp/src/sort.cpp
```diff
@@ -105,6 +331,150 @@ std::unique_ptr<cudf::column> get_split_ind(TaskContext& ctx,
...
+  // Create a column with the same length as sorted table, filled with current rank
+  auto rank_column = create_column<int32_t>(sorted_table.num_rows(), ctx.rank, ctx);
+
+  // Create a new table view by appending the rank column to the sorted table
+  std::vector<cudf::column_view> table_columns;
+  for (int i = 0; i < sorted_table.num_columns(); i++) {
+    table_columns.push_back(sorted_table.column(i));
+  }
+  table_columns.push_back(rank_column->view());
+  auto sorted_table_with_rank = cudf::table_view(table_columns);
```
seberg, line 455: So you append the rank, but not the row-id?  I don't really want to time which strategy is better, so let's keep this and I might open an issue.

FWIW, I suspect we could just use the global row-id at this point if we already add a column.

Either way, I think it's probably better to consider this in a follow up (that is really not urgent).

cpp/src/sort.cpp
```diff
@@ -93,10 +95,234 @@ std::unique_ptr<cudf::column> get_split_ind(TaskContext& ctx,
...
+  std::shared_ptr<arrow::Table> global_split_values,
+  const std::vector<cudf::size_type>& keys_idx,
+  const std::vector<arrow::compute::SortKey>& column_order,
+  arrow::compute::NullPlacement null_precedence)
+{
+  // Create a new table with the rank column appended
+  auto rank_array = create_array<int64_t>(sorted_table->num_rows(), ctx.rank);
+  auto sorted_table_with_rank =
+    ARROW_RESULT(sorted_table->AddColumn(sorted_table->num_columns(),
+                                         arrow::field("split_rank", rank_array->type()),
```
seberg, line 253: I guess we need a name, although, in theory it could collide with an existing one.  Could see taking on this as a small debt.  Making a new name isn't that trivial I guess (unless we just make it longer), since you would have to e.g. use `("_" * max_len) + "split_rank"` or so.

seberg (commented):
> The comparison I have implemented so far by scratch could instead be an arrow sort over two rows

Seems you already did this now, I agree that vectorizing may be interesting, but happy to also defer to not keep iterating here unnecessarily.
Overall, LGTM, I don't think I had any big comments going through (may want to make another brief pass, though).

There are a couple of smaller comments, but e.g. considerations like going towards a `global_idx` row (rather than `rank` row now that we are not using two lower-bound calls), are probably better done in a follow-up.

RAMitchell:
One more issue: NaNs are considered slightly differently in arrow and cudf. Arrow treats them as `None` but less than actual `None`. Cudf treats them as the largest floating point number. I don't think I have any way to resolve this difference between the CPU and GPU launch of sort.

cpp/src/core/repartition_by_hash.cu
```diff
@@ -123,8 +127,111 @@ class ExchangedSizes {
...
+
+  if (ctx.get_legate_context().communicators().empty()) {
+    throw std::runtime_error("internal error: communicator not initialized.");
+  }
+  auto comm      = ctx.get_legate_context().communicator(0);
+  auto* comm_ptr = comm.get<legate::comm::coll::CollComm>();
+
+  // Serialize table as buffers
+  std::vector<std::shared_ptr<arrow::Buffer>> send_buffers;
+  std::vector<int> sendcounts;
```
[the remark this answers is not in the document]
RAMitchell: MPI and legate's implementation of its API accepts int unfortunately.

seberg:
> Cudf treats them as the largest floating point number. I don't think I have any way to resolve this difference between the CPU and GPU launch of sort.

Ack... Yeah, let's not worry about it can't see much more than adding an `xfail()` test or a note in the docs about it.

In practice, I think the only solution we have is to create a `nans_to_null` function and ask the user to call it themselves when needed (or call it for them in polars logic).
But somehow I am not convinced of calling it implicitly here right now (although... I did for `min/max`, but for min/max the result is just wrong without it not just inconsistent).

cpp/src/core/repartition_by_hash.cu
```diff
@@ -123,8 +127,111 @@ class ExchangedSizes {
...
+      ss << "Failed to write table to stream: " << status.ToString();
+      throw std::runtime_error(ss.str());
+    }
+    send_buffers.push_back(ARROW_RESULT(output_stream->Finish()));
+    sendcounts.push_back(send_buffers.back()->size());
+  }
+
+  std::size_t total_send_size =
+    std::accumulate(sendcounts.begin(), sendcounts.end(), 0, std::plus<std::size_t>());
+  auto sendbuffer = legate::create_buffer<uint8_t>(total_send_size);
```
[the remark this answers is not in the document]
seberg: Ahh, hmmm, I think [`Tell()`](https://arrow.apache.org/docs/cpp/api/io.html#_CPPv4N5arrow2io18BufferOutputStreamE) should works, but not 100% sure.

cpp/src/sort.cpp
```diff
@@ -401,27 +719,30 @@ LogicalTable sort(const LogicalTable& tbl,
...
+  argument::add_next_scalar(task, nulls_at_end);
   argument::add_next_scalar(task, stable);
   argument::add_next_output(task, ret);
 
-  task.add_communicator("nccl");
+  if (use_arrow) {
+    task.add_communicator("cpu");
+  } else {
+    task.add_communicator("nccl");
+  }
```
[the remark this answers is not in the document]
seberg: Thanks for pointing that out.  Seems strange to me, but if they use the same pattern currently, definitely no point in fretting over it!

cpp/src/sort.cpp
seberg: Ah, you are right, we would need that.  I am not sure if you can get it locally via `arr.shape<1>().lo[0]`.  I suspect that works (since we have a concurrent task), but I am not actually fully sure.

Either way, seems like a follow up and you are right, I am not quite sure if it's easy.

python/legate_dataframe/lib/sort.pyx
```diff
@@ -33,26 +29,29 @@ def sort(
     LogicalTable tbl,
     list keys,
     *,
-    list column_order = None,
-    list null_precedence = None,
+    list sort_ascending = None,
+    nulls_at_end = True,
     stable = False,
 ):
     """Perform a sort of the table based on the given columns.
```
seberg, line 36:
```suggestion
    """Perform a sort of the table based on the given columns.

```

seberg (approved):
Thanks for the updates, let's go with this and see if we want to iterate (but I think nothing is lost if this is the status quo for a long time, more curious whether e.g. the approach to finding the splitting points matters in practice).

PR #79, opened 2025-07-10
ENH: Allow ingesting arrow large strings and fix timedelta/datetime roundtrip

seberg:
Does not yet use them when our strings are larger than 2GiB. I need this for polars interop because polars tends to create large strings (probably just always).
(So tested there, but not here, but I guess we could add explicit arrow round-trip tests.)

Basically, just duplicating.  Templating didn't seem much better considering how short the snippets were, but if you have an idea, happy to change.

### DIFF
--- a/ci/run_pytests_cpu.sh
+++ b/ci/run_pytests_cpu.sh
@@ -28,5 +28,5 @@ legate \
     . \
     -sv \
     --durations=0 \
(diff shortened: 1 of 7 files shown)


RAMitchell (approved):
Would be good to see at least 1 test.

seberg:
Yeah, let me add a test for arrow array/column round-tripping before putting this in.

PR #88, opened 2025-07-30
Add explicit NCCL depdencency

seberg:
This is largely cargo-culted from raft.  It seems that newer legate versions do not use nccl anymore and our transitive dependency will break.

Not clearly tested without the new legate, so we may want to wait until we actually see the change (which means until cupynumeric updates it's legate dependency).

---

Ping @reazulhoque since you ran into this with the newer legate.

### DIFF
--- a/conda/recipes/legate-dataframe/meta.yaml
+++ b/conda/recipes/legate-dataframe/meta.yaml
@@ -103,7 +103,7 @@ outputs:
         - pynvjitlink<=0.6
     ignore_run_exports_from:
       - cupynumeric  # only a build dependency to help resolver
-      - nncl  # cargo-culted from raft (assume the pin is overly strict)
+      - nccl  # taken from raft (maybe due to being only a major version pin?)
     run_exports:
       - {{ pin_subpackage("legate-dataframe", max_pin="x.x") }}
     test:
--- a/conda/recipes/legate-dataframe/meta.yaml
+++ b/conda/recipes/legate-dataframe/meta.yaml
@@ -60,6 +60,7 @@ requirements:
     - cudf {{ rapids_version }}  # pulls in pyarrow?
     - libcudf {{ rapids_version }}
     - librmm {{ rapids_version }}
+    - nccl {{ nccl_version }}
     - pylibcudf {{ rapids_version }}
 
 
@@ -96,11 +97,13 @@ outputs:
         # released, a new legate-dataframe would be required.
         - legate
         - numpy >=1.23,<3.0.0a0
+        - nccl {{ nccl_version }}
         - rmm {{ rapids_version }}
         - cudf {{ rapids_version }}
         - pynvjitlink<=0.6
     ignore_run_exports_from:
       - cupynumeric  # only a build dependency to help resolver
+      - nncl  # cargo-culted from raft (assume the pin is overly strict)
     run_exports:
       - {{ pin_subpackage("legate-dataframe", max_pin="x.x") }}
     test:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -15,6 +15,7 @@ files:
       - depends_on_legate
       - depends_on_libcudf
       - depends_on_librmm
+      - depends_on_nccl
       - docs
       - py_version
       - run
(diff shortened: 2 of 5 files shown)


seberg:
> Just to clarify I am not trying it with newer Legate explicitly (as in custom Legate)

Ah OK, but I am not clear at all why we don't see the problem in CI then.  If you happen to try this and works for you can you ping us, then I think we should just merge it.

conda/recipes/legate-dataframe/meta.yaml
```diff
@@ -96,11 +97,13 @@ outputs:
         # released, a new legate-dataframe would be required.
         - legate
         - numpy >=1.23,<3.0.0a0
+        - nccl {{ nccl_version }}
         - rmm {{ rapids_version }}
         - cudf {{ rapids_version }}
         - pynvjitlink<=0.6
     ignore_run_exports_from:
       - cupynumeric  # only a build dependency to help resolver
+      - nncl  # cargo-culted from raft (assume the pin is overly strict)
```
jameslamb, line 106:
```suggestion
      - nccl  # cargo-culted from raft (assume the pin is overly strict)
```

This looks like a typo.

jameslamb: > _assume the pin is overly strict_

I'm not sure why `raft` recipes ignore the run exports on `nccl`. The `nccl` from `conda-forge` only has a major-version pin as its run export (https://github.com/conda-forge/nccl-feedstock/blob/d10b348347db1a1175654f10e6c24f5d524c8652/recipe/meta.yaml#L23) which shouldn't be too restrictive. Maybe that `ignore_run_exports` entry is to help in the case where we get `nccl` from the `nvidia` channel 🤷🏻 

I think it's fine if you want to ignore those run exports here too (as this PR is doing). But let's see what CI says once you fix this typo.

dependencies.yaml
```diff
@@ -249,3 +255,19 @@ dependencies:
           - matrix:
             packages:
               - *librmm_unsuffixed
+
+  depends_on_nccl:
+    common:
+      - output_types: conda
+        packages:
+          - &nccl_unsuffixed nccl>=2.19
```
jameslamb, line 263:
```suggestion
          - nccl>=2.19
```

This `nccl_unsuffixed` alias isn't getting used anywhere... it should be removed.

dependencies.yaml
```diff
@@ -165,6 +168,9 @@ dependencies:
       - output_types: [conda, pyproject, requirements]
         packages:
           - numpy >=1.23,<3.0.0a0
+      - output_types: conda
+        packages:
+          - nccl
```
jameslamb, line 174:
```suggestion
```

I think this should be removed. It could cover up mistakes in `legate-dataframe`'s packaging. We should rely on the `legate-dataframe` conda packages' `run:` dependency on `nccl` to pull this in.

conda/recipes/legate-dataframe/meta.yaml
```diff
@@ -60,6 +60,7 @@ requirements:
     - cudf {{ rapids_version }}  # pulls in pyarrow?
     - libcudf {{ rapids_version }}
     - librmm {{ rapids_version }}
+    - nccl {{ nccl_version }}
```
jameslamb, line 63: Yeah this is a good fix! The existence of direct references like this:

https://github.com/rapidsai/legate-dataframe/blob/49cc2b9fe586e1144208bad7d7ceddfde09ac6d5/cpp/include/legate_dataframe/core/nccl.hpp#L21

mean we should have a `host:` dependency on `nccl` here.

jameslamb (approved):
Based on the description of the PR, I'm supportive of this change. I left some small suggestions, but I don't need to re-review after you've addressed them.

`@` me if you have questions or do want another review.

conda/recipes/legate-dataframe/meta.yaml
seberg: Should have checked, I'll remove the comment. Probably it's the opposite then: we want a newer NCCL, but the run-export doesn't ensure that.

jameslamb: Ok yeah, makes sense.

PR #92, opened 2025-08-01
Implement cpu join

RAMitchell:


cpp/src/join.cpp
```diff
@@ -200,8 +204,185 @@ bool is_repartition_not_needed(const TaskContext& ctx,
...
+  arrow::acero::Declaration left{"table_source", arrow::acero::TableSourceNodeOptions{lhs}};
+  arrow::acero::Declaration right{"table_source", arrow::acero::TableSourceNodeOptions{rhs}};
+  arrow::acero::Declaration hashjoin{"hashjoin", {left, right}, std::move(join_opts)};
+
+  return ARROW_RESULT(arrow::acero::DeclarationToTable(std::move(hashjoin)));
+}
+
+// Creating some constant for each arrow type is surprisingly hard
+std::shared_ptr<arrow::Scalar> create_default_value_scalar(
+  const std::shared_ptr<arrow::DataType>& type)
```
seberg, line 278: Feels like we should be able to do this with an overload or template?

cpp/src/join.cpp
```diff
@@ -200,8 +204,185 @@ bool is_repartition_not_needed(const TaskContext& ctx,
...
+                                       key_replaced_nulls));
+    left_fields.emplace_back(arrow::FieldRef(key + "_replaced"));
+  }
+
+  std::vector<arrow::FieldRef> right_fields;
+  for (const auto& key : rhs_keys) {
+    auto key_array = rhs->GetColumnByName(key);
+    auto scalar    = create_default_value_scalar(key_array->type());
+    auto key_replaced_nulls =
+      ARROW_RESULT(arrow::compute::CallFunction("coalesce", {key_array, scalar})).chunked_array();
```
seberg, line 322: I guess it's the same, but let's use:
```suggestion
      ARROW_RESULT(arrow::compute::CallFunction("fill_null", {key_array, scalar})).chunked_array();
```

cpp/src/join.cpp
```diff
@@ -200,8 +204,185 @@ bool is_repartition_not_needed(const TaskContext& ctx,
...
+    lhs_temp =
+      ARROW_RESULT(lhs_temp->AddColumn(lhs_temp->num_columns(),
+                                       arrow::field(key + "_replaced", key_replaced_nulls->type()),
+                                       key_replaced_nulls));
+    left_fields.emplace_back(arrow::FieldRef(key + "_replaced"));
+  }
+
+  std::vector<arrow::FieldRef> right_fields;
+  for (const auto& key : rhs_keys) {
+    auto key_array = rhs->GetColumnByName(key);
```
seberg, line 319: I do feel an `if (!nullable)` is probably sensible here.

cpp/src/join.cpp
```diff
@@ -200,8 +204,185 @@ bool is_repartition_not_needed(const TaskContext& ctx,
...
+    auto scalar    = create_default_value_scalar(key_array->type());
+    auto key_replaced_nulls =
+      ARROW_RESULT(arrow::compute::CallFunction("coalesce", {key_array, scalar})).chunked_array();
+    rhs_temp =
+      ARROW_RESULT(rhs_temp->AddColumn(rhs_temp->num_columns(),
+                                       arrow::field(key + "_replaced", key_replaced_nulls->type()),
+                                       key_replaced_nulls));
+    right_fields.emplace_back(arrow::FieldRef(key + "_replaced"));
```
seberg, line 327: This doesn't seem right to me? But I guess `compare_nulls` isn't properly tested.

There is no safe unique `scalar` that is definitely not the same as any other value. So I think this works, but you additionally have to add the nullmask as another join column to distinguish that case.

cpp/src/join.cpp
```diff
@@ -200,8 +204,185 @@ bool is_repartition_not_needed(const TaskContext& ctx,
...
+  }
+
+  arrow::acero::HashJoinNodeOptions join_opts(legate_to_arrow_join_type(join_type),
+                                              left_fields,
+                                              right_fields,
+                                              arrow::compute::literal(true),
+                                              "_left",
+                                              "_right");
```
seberg, line 335: It seems `HashJoinNodeOptions` supports listing [output keys](https://arrow.apache.org/docs/cpp/api/acero.html#_CPPv4N5arrow5acero19HashJoinNodeOptions19HashJoinNodeOptionsE8JoinTypeNSt6vectorI8FieldRefEENSt6vectorI8FieldRefEENSt6vectorI8FieldRefEENSt6vectorI8FieldRefEE10ExpressionNSt6stringENSt6stringEb) similar to how legate-dataframe passes it.  I think we should probably use that.

seberg (commented):
Mostly comments for myself right now, I'll probably push an update.  With those fixed and some Python tests it does look reasonable though (even if it's rather unfortunate we have to do this dance for null matching).

python/tests/test_join.py
```diff
@@ -126,28 +154,28 @@ def test_column_names():
         lhs_out_columns=["data0", "key"],
         rhs_out_columns=["data1"],
     )
-    assert_frame_equal(
-        res,
-        cudf.DataFrame({"data0": [1, 2, 3], "key": [1, 2, 3], "data1": [3, 2, 1]}),
-        ignore_row_order=True,
```
seberg, line 132: I think this test is just failing because the result has a different order and it needs to sort by the key column (fun that the gpu one passes in CI).

seberg (approved):
Thanks for improving the python tests as well!  I am happy with this, I applied all the changes that I cared about so if those look fine to you, we are good.

PR #93, opened 2025-08-06
DO NOT MERGE: Run CI on "unmodified" main

seberg:
After legate version bump, CI seems to segfault.  I could not reproduce locally (neither with a local build, nor with CI build artifacts).

So confirm once here, and maybe I/we have an idea to try other things in CI.


seberg:
OK, so the problem isn't legate or nvtx and the other changes seem pretty irrelevant.

The likely change is:  The runners were updated to use CUDA 13.0 drivers and that is the change causing chaos here, it just happened basically at the same time as the legate 25.08.00.rc1.

seberg:
OK, so the solution that seems to work is to patch numba-cuda until we can update cudf within CI easily, because that pins to an old numba-cuda version which breaks with the new driver versions.

This only affects us indirectly within tests (we don't use any of this directly here).

ci/numba_cuda.diff
```diff
@@ -0,0 +1,14 @@
...
+@@ -365,6 +365,9 @@ class Driver(object):
+         else:
+             variants = ('_v2', '')
+
++        if fname in ("cuCtxGetDevice", "cuCtxSynchronize"):
++            return getattr(self.lib, fname)
++
+         for variant in variants:
+             try:
+                 return getattr(self.lib, f'{fname}{variant}')
```
seberg, line 14: Just for reference, the fix comes from: https://github.com/NVIDIA/numba-cuda/pull/185/files (as per Graham, this being the likely fix -- and it was :)).

seberg:
Ah, dang, should have thought of that :facepalm:. I'll try with the earliest and go back only if that creates an issue.

jameslamb:
🎉 looks like that worked!

PR #94, opened 2025-08-07
ENH: Inform legate that some joins don't need a concurrent task launch

seberg:
If we don't need a communicator because we have a broadcast join, we also don't need a concurrent task launch at all.

Includes gh-95 to make the templating slightly nicer.

---

Didn't try locally (machine I usually use is currently unavailable).


seberg:
Reazul tried this and it didn't work.  Legion doesn't like this being changed on the task.  The opposite direction probably works, but warns.
W'll have to subclass and just have two tasks versions here, that seems fine to me. This use-case is probably not very common.

RAMitchell (approved):

PR #97, opened 2025-08-11
Refactor cuda code into .cu files

RAMitchell:
Start with binaryop and readcsv. Split into .cu and .cpp files to prepare for cpu only compilation. Also change cudf data types to arrow data types in function signature.

### DIFF
--- a/cpp/include/legate_dataframe/core/column.hpp
+++ b/cpp/include/legate_dataframe/core/column.hpp
@@ -82,6 +82,31 @@ class LogicalColumn {
     }
   }
 
+  /**
+   * @brief Create a column with a legate array as the data
+   *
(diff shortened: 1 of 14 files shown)


cpp/include/legate_dataframe/core/column.hpp
```diff
@@ -82,6 +82,31 @@ class LogicalColumn {
     }
   }
 
+  /**
+   * @brief Create a column with a legate array as the data
```
RAMitchell, line 86: This duplicates the cudf data type version, which I expect to remove in the future.

seberg:
Ack, sorry brain-slip... you are still explicitly supporting cudf dtypes, so clearly on the Python side there isn't a breaking change yet, just a preference change.

PR #100, opened 2025-08-14
Expose datetime component to polars and fix true-divide

seberg:
We should still expose datetime component extraction for CPU as well probably, that might change things, but hopefully too annoyingly so.

I think true-divide (at least for polars) is main useful thing, while divide I believe is a floor-divide, which we could expose but only easily for non-floats probably.
So, hacked true-divide to work, although one day we may want to lower that.

### DIFF
--- a/cpp/src/binaryop.cu
+++ b/cpp/src/binaryop.cu
@@ -37,6 +37,7 @@ cudf::binary_operator arrow_to_cudf_binary_op(std::string op, legate::Type outpu
   // https://docs.rapids.ai/api/libcudf/stable/group__transformation__binaryops
   std::unordered_map<std::string, cudf::binary_operator> arrow_to_cudf_ops = {
     {"add", cudf::binary_operator::ADD},
+    // NOTE: if we enable true divide here, should improve polars side.
     {"divide", cudf::binary_operator::DIV},
     {"multiply", cudf::binary_operator::MUL},
     {"power", cudf::binary_operator::POW},
--- a/cpp/src/parquet.cpp
+++ b/cpp/src/parquet.cpp
@@ -250,6 +250,9 @@ class ParquetRead : public Task<ParquetRead, OpCode::ParquetRead> {
     auto opt = cudf::io::parquet_reader_options::builder(src);
     opt.columns(columns);
     opt.row_groups(row_groups);
+    // If pandas metadata is read, libcudf may read index columns without this.
+    opt.use_pandas_metadata(false);
+
     auto res = cudf::io::read_parquet(opt, ctx.stream(), ctx.mr()).tbl;
 
     if (get_prefer_eager_allocations()) {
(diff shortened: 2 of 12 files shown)


python/legate_dataframe/ldf_polars/__init__.py
```diff
@@ -45,7 +45,7 @@ def lazy_from_legate_df(df):
 
     # Last False (within polars) changes the predicate to be passed for arrow.
     plf = PyLazyFrame.scan_from_python_function_pl_schema(
-        schema, partial(_create_df, df), False
+        schema, partial(_create_df, df), False, validate_schema=True
```
seberg, line 48: Seems to have bit-rot.  Hopefully, it works on most polars versions (added a test).

python/legate_dataframe/ldf_polars/dsl/expressions/binaryop.py
```diff
@@ -89,6 +89,17 @@ def do_evaluate(
...
+            return Column(binary_operation(left.obj, right.obj, self.op, self.dtype))
+
+        # Use divide, but cast one of the inputs to the output dtype
+        # to ensure the right type is used (with some logic to prefer
+        # casting a scalar).
+        if left.obj.type() != self.dtype and right.obj.type() != self.dtype:
+            if left.is_scalar:
+                left = left.astype(self.dtype)
+            else:
+                right = right.astype(self.dtype)
```
seberg, line 103: I figured that the result dtype should be a correct float and casting the input is fine and probably what would happen (internally thought).
But thought to skip if clearly not needed and to prefer casting the scalar side if we have one.

python/legate_dataframe/ldf_polars/dsl/expressions/datetime.py
```diff
@@ -133,93 +134,71 @@ def do_evaluate(
...
-                plc.types.DataType(plc.types.TypeId.INT32),
+                "add",
+                pa.int32(),
             )
             return Column(total_nanos)
 
         return Column(
-            plc.datetime.extract_datetime_component(
+            timestamps.extract_timestamp_component(
                 column.obj,
```
seberg, line 202: Forgot to exclude this file, but the diff is nice maybe ;).

cpp/src/parquet.cpp
```diff
@@ -250,6 +250,9 @@ class ParquetRead : public Task<ParquetRead, OpCode::ParquetRead> {
     auto opt = cudf::io::parquet_reader_options::builder(src);
     opt.columns(columns);
     opt.row_groups(row_groups);
+    // If pandas metadata is read, libcudf may read index columns without this.
+    opt.use_pandas_metadata(false);
```
seberg, line 254: Ran into this with a weirder pandas generated parquet file, would need to include the parquet file to actually test (since I don't want to generate weird pandas metadata).

cpp/src/binaryop.cu
```diff
@@ -37,6 +37,7 @@ cudf::binary_operator arrow_to_cudf_binary_op(std::string op, legate::Type outpu
   // https://docs.rapids.ai/api/libcudf/stable/group__transformation__binaryops
   std::unordered_map<std::string, cudf::binary_operator> arrow_to_cudf_ops = {
     {"add", cudf::binary_operator::ADD},
+    // NOTE: if we enable true divide here, should improve polars side.
```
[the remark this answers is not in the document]
RAMitchell: Its fine as is.

PR #104, opened 2025-08-18
Implement cpu timestamps

RAMitchell:


cpp/src/timestamps.cpp
```diff
@@ -91,21 +91,20 @@ LogicalColumn to_timestamps(const LogicalColumn& input,
...
 }
 
-LogicalColumn extract_timestamp_component(const LogicalColumn& input,
-                                          cudf::datetime::datetime_component component)
+LogicalColumn extract_timestamp_component(const LogicalColumn& input, std::string component)
 {
-  if (!cudf::is_timestamp(input.cudf_type())) {
+  auto timestamp_type = std::dynamic_pointer_cast<arrow::TimestampType>(input.arrow_type());
+  if (!timestamp_type) {
     throw std::invalid_argument("extract_timestamp_component() input must be timestamp");
```
seberg, line 98: Hmmm, I guess most of this will already throw in the `std::dynamic_pointer_cast` and this just checks whether the `shared_ptr` points to something.
Should we add a test? (the old error is a bit different, but doesn't really matter.)

cpp/src/timestamps.cu
```diff
@@ -0,0 +1,88 @@
...
+  const auto component = argument::get_next_scalar<std::string>(ctx);
+  const auto input     = argument::get_next_input<PhysicalColumn>(ctx);
+  auto output          = argument::get_next_output<PhysicalColumn>(ctx);
+
+  std::unique_ptr<cudf::column> ret;
+  ret = cudf::datetime::extract_datetime_component(
+    input.column_view(), to_cudf_component(component), ctx.stream(), ctx.mr());
+
+  // Cast to int64
+  ret = cudf::cast(ret->view(), cudf::data_type(cudf::type_id::INT64), ctx.stream(), ctx.mr());
```
seberg, line 79: Hmmm, so we change the result type to always be int64.

cpp/src/timestamps.cpp
RAMitchell: Good idea, will add a test.

cpp/src/timestamps.cu
seberg: Yeah, that's true, from a Python API perspective, I like just using 64bit integers.

PR #105, opened 2025-08-20
Express column type in terms of arrow instead of cudf

RAMitchell:


cpp/src/utils.cpp
```diff
@@ -305,4 +354,42 @@ size_t linearize(const legate::DomainPoint& lo,
...
+  const std::vector<uint8_t>& data)
+{
+  auto buffer       = arrow::Buffer::Wrap(data.data(), data.size());
+  auto input_stream = arrow::io::BufferReader(buffer);
+  auto schema       = ARROW_RESULT(arrow::ipc::ReadSchema(&input_stream, nullptr));
+  std::vector<std::shared_ptr<arrow::DataType>> types;
+  for (const auto& field : schema->fields()) {
+    types.push_back(field->type());
+  }
+  return types;
```
seberg, line 393: I really like this serialization!

However, one (hopefully small) ask is that I think it would be nice to add overloads to `task_argument.cpp/hpp` (some stuff there are inline versions, so for simplicity maybe `.hpp`).

Basically, I would like to be able to write:
```
auto arrow_type = argument::get_next_scalar<arrow::DataType>(ctx));
```
(or `std::shared_ptr<arrow::DataType>`) and also the same for `argument::add_next_scalar(arrow_type)` to just work.

The code here can stay, I just think it would be nice to not have to call the serialize/deserialize explicitly anyway.

seberg (approved):
Nice, I'll trust the column code move is mostly a plain move, cool to see that split into `.cpp`/`.cu`.

I do like how the serialization and use of arrow turned out!  One thing I feel would be nice is to just support the arrow type directly in the `get_next_scalar`, etc. functions.

We/I could do that in a follow-up also to remove the serialize/deserialize, though.

PR #108, opened 2025-08-25
Try keeping image partition array alive

seberg:
This is a work-around and should probably not be merged.

Legate has some problems when image constraints get lost, this is probably uninteresting as soon as there is better support for unbound stores in streaming (my guess only!) because at that point the change of partitions shouldn't matter.

Legate/legion basically gets confused how to partition the store and repartitions it unnecessarily breaking some of the analysis leading to non-ideal usage.


seberg:
Closing this, this was never meant to be merged and I think the branch is now unnecessary for the use-case.

PR #116, opened 2025-08-29
Allow groupby count-valid with and without nulls and expose to polars

seberg:
This channels through the FunctionOptions from arrow all the way into the task, the nice thing is that those come already with serialization capabilities.

The only thing really supported right now (on the GPU) is for counting NULLs or not.  We could support more in the future.

We might add more options in the future (but maybe also disable CPU ones if they are not amendable to improving the distributed algorithm).

## Checklist
- [x] Run `./build.sh test` for local testing, see [CONTRIBUTING.md](https://github.com/rapidsai/legate-dataframe/blob/main/CONTRIBUTING.md#testing).

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -75,8 +75,7 @@ rapids_cpm_init()
 
 # If CCCL versions don't get picked up correctly, we may need to add them early on here (right now
 # it seems fetching rmm/cudf manage to ensure we ge the right one.)
-include("${rapids-cmake-dir}/cpm/cccl.cmake")
-rapids_cpm_cccl()
+# include("${rapids-cmake-dir}/cpm/cccl.cmake") rapids_cpm_cccl()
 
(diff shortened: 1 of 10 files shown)


cpp/CMakeLists.txt
```diff
@@ -75,8 +75,7 @@ rapids_cpm_init()
 
 # If CCCL versions don't get picked up correctly, we may need to add them early on here (right now
 # it seems fetching rmm/cudf manage to ensure we ge the right one.)
-include("${rapids-cmake-dir}/cpm/cccl.cmake")
-rapids_cpm_cccl()
+# include("${rapids-cmake-dir}/cpm/cccl.cmake") rapids_cpm_cccl()
```
seberg, line 78:
```suggestion
include("${rapids-cmake-dir}/cpm/cccl.cmake")
rapids_cpm_cccl()
```
whoops, sorry.

PR #120, opened 2025-09-02
ENH: Add contains and polars is_in

seberg:
Unfortunately, polars relies on list dtype here and we don't support that yet, so there is a bit of a hack to rewrite the literal part of the query to use the needed dtype and not the list one. (I hope this is OK in practice and assume it can't really go wrong badly although it can imagine it failing.)

Otherwise, relatively straight forward `contains` added in a `search.hpp`, etc. the polars code is mostly cudf-polars boilerplate.

---

Draft, need to rebase after the other PR (to avoid test failures due to `test_task.py`).

### DIFF
--- a/cpp/include/legate_dataframe/search.hpp
+++ b/cpp/include/legate_dataframe/search.hpp
@@ -24,7 +24,9 @@ namespace task {
 class ContainsTask : public Task<ContainsTask, OpCode::Contains> {
  public:
   static void cpu_variant(legate::TaskContext context);
+#ifdef LEGATE_DATAFRAME_USE_CUDA
   static void gpu_variant(legate::TaskContext context);
+#endif
 };
 
 }  // namespace task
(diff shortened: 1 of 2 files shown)


cpp/include/legate_dataframe/search.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <legate_dataframe/core/library.hpp>
+
+namespace legate::dataframe {
+
+namespace task {
+
+class ContainsTask : public Task<ContainsTask, OpCode::Contains> {
+ public:
+  static void cpu_variant(legate::TaskContext context);
+  static void gpu_variant(legate::TaskContext context);
```
RAMitchell, line 27: Would appreciate if you can mask the gpu variant in the macro even though its not tested in CI yet.

RAMitchell (approved):

cpp/include/legate_dataframe/search.hpp
seberg: Good catch, I forgot. I think also probably in the copying will add it here for both.

PR #123, opened 2025-09-03
ENH: Enable boolean unary operations

seberg:
This enables boolean unary operations. It isn't quite clear to me yet that we should integrate them into `unary_operation`. Maybe it would make sense to split out these that always return booleans (also from a typing perspective they are just a bit different). OTOH, this isn't a convenient entry-point either way...


seberg:
Pretty simple, but more of an API question whether we should split these out from `unary_operation` (not that we are beholden to it quite yet, but still).

RAMitchell (approved):
LGTM!

PR #126, opened 2025-09-04
ENH: Allow `chunk_by_rows=True` for parquet reading

seberg:
Allows to not read by row-group granularity and side-steps the image constraints for the eager-alloc preference variable (In python mode).
Option is always available though.

The new code fails for CPU strings (although only after the read is done).  Something is invalid there, it isn't clear to me yet what or why.
(Do recordbatches not slice gracefully for strings or so?)


cpp/src/core/ranges.cpp
```diff
@@ -42,18 +42,34 @@ std::shared_ptr<arrow::Buffer> global_ranges_to_arrow_offsets(const legate::Phys
...
-  for (size_t i = 0; i < array.length(); ++i) {
-    ranges_acc[i].lo[0] = array.value_offset(i);
-    ranges_acc[i].hi[0] = array.value_offset(i + 1) - 1;
+  auto first_offset = array.value_offset(0);
+  if (first_offset == 0) {
+    for (size_t i = 0; i < array.length(); ++i) {
+      ranges_acc[i].lo[0] = array.value_offset(i);
+      ranges_acc[i].hi[0] = array.value_offset(i + 1) - 1;
+    }
+  } else {
```
RAMitchell, line 51: I think you only need the second branch here.

cpp/src/core/ranges.cpp
```diff
@@ -42,18 +42,34 @@ std::shared_ptr<arrow::Buffer> global_ranges_to_arrow_offsets(const legate::Phys
...
 }
 
 void arrow_offsets_to_local_ranges(const arrow::LargeStringArray& array,
                                    legate::Rect<1>* ranges_acc)
 {
-  for (size_t i = 0; i < array.length(); ++i) {
-    ranges_acc[i].lo[0] = array.value_offset(i);
-    ranges_acc[i].hi[0] = array.value_offset(i + 1) - 1;
+  auto first_offset = array.value_offset(0);
+  if (first_offset == 0) {
```
RAMitchell, line 63: Same here.

python/legate_dataframe/lib/parquet.pyx
```diff
@@ -80,6 +86,13 @@ def parquet_read(files, *, columns=None) -> LogicalTable:
         otherwise must be a path or an iterable of paths (or strings).
     columns
         List of strings selecting a subset of columns to read.
+    chunk_by_rows
+        If set to ``True`` the read operation will not be chunked into row groups.
```
RAMitchell, line 90: This deserves perhaps a little more explanation. Also the name now looks a little weird to me in that chunk_by_rows means don't use row_groups. Maybe ignore_row_groups?

I.e. Parquet files may arrive with defined row_groups - by default we use these row groups to enable parallelism, where each legate worker can process one or more row groups. Sometimes a parquet arrives with very large row groups (in the order of 10GB or more). This limits us, particularly on the CPU, if we can only launch 1 thread per 10GB of data. Setting this to true ignores metadata defined row_groups and creates a custom partitioning.

RAMitchell (approved):

cpp/src/core/ranges.cpp
RAMitchell: Its no big deal just looks like classic premature optimisation.
https://wiki.c2.com/?PrematureOptimization=

python/legate_dataframe/lib/parquet.pyx
seberg: Let's go with `ignore_row_groups`, I don't like either name much TBH (also it's difficult to actually say what the trade-offs/resource use is exactly).

I.e. on the CPU, we are still limited with this... since arrow seems to have no way to skip rows quickly (e.g. without full decompression).

I am wondering if there is a predicate API that can do this (I think there is a predicate API, but I am not sure if it has a row number logic).

PR #135, opened 2025-09-17
Skip files without rows in rowwise cuda parquet read

seberg:
This avoids touching those files which is somewhat nice.

Otherwise, it works around an issue with int32 overflows in older cudf versions that is presuambly already fixed (it doesn't fully fix it).

On newer cudf versions, we may want to re-organize things to a chunked reader.  Or maybe the newer one even supports row-groups + row offset (I am not sure) which would make it maybe slightly nicer.

Unfortunately, hard to excercise, because you need multiple files that with a combined size surpassing 2GiB (at least after decompression).


RAMitchell (approved):
LGTM!

PR #139, opened 2025-09-25
Use `large_string` always and change back list unpacking logic

seberg:
Mixing string and `large_string` seemsm very annoying, since it breaks `==` comparisons which would have to be carefull replaced everywhere. Since I doubt there is a large downside to large strings and legate technically is large string (64bit), decided to normalize to that for now to side-step having to change the comparisons.

But... I could go either way, as it is it just created problems (to be fair in part this might only work out because polars also uses `large_string` always).

---

I am seeing more hangs than I used to locally right now I feel, but not sure why if there is any particular reason for that (i.e. newer legate hanging more often)...

### DIFF
--- a/cpp/include/legate_dataframe/core/ranges.hpp
+++ b/cpp/include/legate_dataframe/core/ranges.hpp
@@ -46,6 +46,14 @@ namespace legate::dataframe {
   rmm::mr::device_memory_resource* mr);
 #endif
 
+/**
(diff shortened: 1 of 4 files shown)


seberg:
Well, over-optimistic here, that seems to confuse things a lot.  But also the way the GPU tests fail looks a bit scary because it seems to use the CPU?

RAMitchell (approved):

PR #142, opened 2025-09-29
Add basic distinct (unique) capability

seberg:
This adds distinct (unique in polars/pandas, but libcudf call is distinct and uses unique for the same operation on pre-sorted arrays).

Arrow seems not to have such a thing, one might be able to do it via structured array columns as well, but groupby and unique are largely the same operation, so I opted to implement it via groupby.

It seemed logical to optimize this for few results (i.e. many dropped columns) by minimizing how much we shuffle, but added an option to not do that.  (If legate-df ever does cardinality estimation via metadata, that could be used for a heuristic here).

### DIFF
--- a/cpp/include/legate_dataframe/stream_compaction.hpp
+++ b/cpp/include/legate_dataframe/stream_compaction.hpp
@@ -30,6 +30,20 @@ class ApplyBooleanMaskTask : public Task<ApplyBooleanMaskTask, OpCode::ApplyBool
 #endif
 };
 
+class DistinctTask : public Task<DistinctTask, OpCode::Distinct> {
+ public:
+  static constexpr auto CPU_VARIANT_OPTIONS =
+    legate::VariantOptions{}.with_has_allocations(true).with_concurrent(true);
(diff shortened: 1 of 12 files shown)


cpp/include/legate_dataframe/stream_compaction.hpp
```diff
@@ -44,4 +58,21 @@ class ApplyBooleanMaskTask : public Task<ApplyBooleanMaskTask, OpCode::ApplyBool
...
+/**
+ * @brief Find distinct (unique) rows in a logical table.
+ *
+ * For rows not listed as keys any value is possible, order is not guaranteed.
+ *
+ * @param tbl The table to distinct.
+ * @param keys The column names to distinct by.
+ * @param high_cardinality Whether the table is assumed to have a high cardinality.
+ * If false (default), the cardinality is assumed to be low in which case we can save
+ * communication by doing a local distinct before shuffling data.  If the cardinality is high,
```
RAMitchell, line 70: I think the default strategy is much more robust and you might not need the paramenter. Have you tested any performance improvement for high cardinality inputs?

Issue #143, opened 2025-09-29
ApplyBooleanMaskTask` fails with "CUDF failure: Column size mismatch" during large GPU benchmark


eginez:
Hi all
I am running into an issue with the [apply_boolean_mask](https://rapidsai.github.io/legate-dataframe/api/table_funcs.html#legate_dataframe.lib.stream_compaction.apply_boolean_mask) function

When using Legate DataFrame (legate version 25.8.0, Legion version 25.8.0) to repeatedly apply boolean masks to large DataFrames on GPU, I encounter a runtime failure:
```
Task "legate::dataframe::task::ApplyBooleanMaskTask" threw an unexpected exception (cudf::logic_error): "CUDF failure at: .../apply_boolean_mask.cu:75: Column size mismatch".
```

#### Reproduction Scenario
- I am running with ` legate --gpus 2`
- I am running a workload that constructs large DataFrames (100,000+ rows) and repeatedly applies boolean masks to filter rows.
- The boolean mask is always constructed to be the same length as the DataFrame it is applied to.
- After several successful mask applications, the process aborts with the above error.

#### Observed Behavior

- The error occurs after several successful mask applications, not immediately.
-  The size of the mask and the dataframe  are indeed the same
- The stack trace points to `ApplyBooleanMaskTask` in legate_dataframe, which wraps a cuDF call.
- The error message is: `"CUDF failure at: .../apply_boolean_mask.cu:75: Column size mismatch"`

#### System Info

- OS: Ubuntu 22.04, Linux 6.8.0-60-generic
- legate libs
```
cupynumeric               25.08.00        cuda12_py311_gpu_g7146e78b_0    legate
legate                    25.08.00        cuda12_py311_ucx_gpu_g47132dac_0    legate
legate-dataframe          25.08.00         cuda12_py311_0    legate
ucc                       1.4.4                h8e98bc0_1    legate
```
-  nvidia-smi
```
Mon Sep 29 16:42:22 2025
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.230.02             Driver Version: 535.230.02   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
[...]

PR #144, opened 2025-09-30
BUG: Ensure alignment in filtering operation

seberg:
Apply boolean mask was missing the alignment between inputs, I had a quick look through, and I think that was the only task.

(Only for the env variable use to use eager allocs, output alignment was also missing for the unary task so added that as well.)

Closes gh-143


seberg:
Test failure doesn't seem to be due to the `unaryop` change (which would have been strange, since the change does nothing in CI).  Instead, it seems we are now running into an assert, maybe the only recent change is even that we are picking up a legate version that _includes_ this assert (but in that case, the assert would be at least overly strict maybe).

RAMitchell (approved):

PR #145, opened 2025-10-02
MAINT: Bump rapids to 25.08 and some other versions

seberg:
This is just a start/try.  CI will pick up a legate with an older CCCL so this may or may not go through.
Locally, I still have problems around `__half` (as typical, I guess), but without cupynumeric, so I should have one with CCCL/rapids-cmake already bumped...


RAMitchell (approved):

seberg:
Ack, this had worked locally with maybe a slightly older legate... Well, let's see if I can mess with some other definition, heh...

seberg:
OK, some of the updates here aren't strictly necessary, but I think they don't hurt either and things are passing.

I have some bad PTX build locally (not sure why), but I don't think that should be a problem (although it means that the new legate isn't tested with multi-gpu setup yet).

PR #150, opened 2025-10-24
Update legate-nightly channel

reazulhoque:
Updating legate nightly conda channel as it has been recently moved.


reazulhoque:
I thought the conda channels in the CI scripts decides where we're putting the `legate-dataframe` package. I am not intending to change that. I am intending to change the source of `Legate` we use to build `legate-dataframe`. Please correct me if I am wrong.

seberg:
Ah no, those are probably the more important ones in the end. The upload channels are also somewhere in the ci scripts/workflow, but I guess it the upload scripts use a different naming convention for the label, so doesn't even show up as `legate/label/...`

EDIT: The ones you find, are all setup for installing packages.  Since they include (except the build one) the special channel for the CI build packages, we don't need to keep `legate/label/experimental` for our own packages there also.

seberg:
Can we already revert the terrible hack from https://github.com/rapidsai/legate-dataframe/pull/145?  The build failure looks like the legate version with the fix might be uploaded (with available cupynumeric, which I hand't expected).

reazulhoque:
I removed the legate hack, CI is failing at a different spot. @Jacobfaib  can you please provide some guidance?

cc: @seberg

seberg:
That error was why the hack existed.  It seems we are in a spot in time where my hack started not working anymore (because the work-around in legate to make it easier is gone?!), but cupynumeric isn't using an updated legate yet so we don't have the fix.
(Unless the legate fix wasn't quite enough.)

seberg:
The CI targets a specific `25.10.*` version always, so yes if we are on 11 now then all 25.10 needs to be changed to 25.11.  I had not realized the nightlies are on 25.11 now.

(I guess ideally we change `VERSION` in a prior commit, just to have a neat commit to tag, but otherwise, I guess can just tag the current main and put this in with the change)

reazulhoque:
@seberg I will need to update the version of legate and cupynumeric. Planning to do it in this PR, does that sound correct?

seberg:
Yes, just search for all 25.10 and update it (after a brief sanity check since e.g. comments should remain correct).

manopapad (approved):

PR #156, opened 2025-11-21
Update legate and cupynumeric to 25.12

reazulhoque:


seberg (approved):
Thanks for updating and bumping versions!

Looks good, *but* you have to still update the `conda/recipe`, though (in a sense that is the more important one, this is mainly to set up a dev environment).
Unfortunately, that is independent from the `dependencies.yaml` setup.

reazulhoque:
can you please point me to that file? I thought these changes are good enough to get a new legate-dataframe package built with `25.12` legate dependencies? Any past PR I can take a look?

